PR #270, opened 2014-05-05
Update Plotly link to Plotly's new user guide

chriddyp:


rgbkrk:
It occurred to me that the plotly notebook is just a listing of links to plotly's website, rather than interesting notebooks on nbviewer.

PR #378, opened 2014-12-10
Bower for most static assets

bollwyvl:
As per #359, this removes all checked-in static assets, and updates `bower.json` (and `.travis.yml`).

This hasn't yet added a submodule of the dependencies: I wasn't sure if that was an approach that was wanted. Do we want to check in `nbviewer/static/components` as a submodule? If so, in this repo (on some permanently-maintained branch), or another repo?

Related to this, the `Dockerfile`, I'm not much a a docker expert. Seems there are several options:
- submodule, as above
- install the whole `node/npm/bower` toolchain into the container 
- second container, do the provision there, then copy it over

Finally, depending on how much we're committing to having asset-oriented toolchain, this could then also be the time to add builds... though this might be yet another PR. Anyhow, LESS and requirejs would suggest themselves, I suppose, and significantly cut down the number of served requests.


bollwyvl:
@rgbkrk I've been dinking around some more with docker: would it be worthwhile to use some of the "best practices" guidance floating around out there to streamline the image a bit? here is what I came up with:
https://gist.github.com/bollwyvl/ff1ed3a83e70b7ba3357

One change that noticably improves its usability during development is adding `requirements.txt` independently of `.` to isolate the changes that could cause a `pip install`. bower can work similarly (though is picker about working dir). unfortunately, one must choose an explicit ordering of `bower` and `pip`, as they are pretty indpendent... i've chosen pip first, as bower should be lighter-weight (no compilation).

bollwyvl:
Again, all of this is selfishly motivated by having an easy-to run hosted instance of nbviewer.ipython.org with all the bells and whistles... which is probably bigger than this PR, but since we're on the topic...

there is mention in `app.py` of using linked docker containers to connect to memcache and elasticsearch... is there any documentation of which appliances are being used? perhaps a `fig.yml` with a sensible setup would help?

Also, looks like combining the `apt-get` stuff even more, and cleaning up after it, might be beneficial in terms of layer complexity...
```
RUN apt-get update \
    && apt-get install -y \
        language-pack-en \
    && locale-gen en_US.UTF-8 \
    && dpkg-reconfigure locales \
    && apt-get install -y \
        build-essential \
        gcc \
        git \
        libcurl4-openssl-dev \
        libevent-dev \
        libmemcached-dev \
        libsqlite3-dev \
        libzmq3-dev \
        make \
        nodejs \
        nodejs-legacy \
        npm \
        pandoc \
        python \
        python-dev \
        python-pip \
        sqlite3 \
        zlib1g-dev \
    && rm -rf /var/lib/apt/lists/*
```

Carreau:
> Sorry for the delay: had a stretch of travel, and then my personal machine died. Back up and running, and as a bonus can now run Docker :)

Hey, coming back after a bit of downtime for end of 2014. 

Sorry to hear about your machine loss, it is always a pain, and I still remember having a machine stolen in a middle of a big PR a few years back, it still hurts. 

Also no need to appologize for the delay, we should be thanking you for contributing ! 
We get back on reviewing shortly (lots of backlog)

rgbkrk:
I'm certainly open to Docker changes. This was the first meaningful Dockerfile I put together and put in production. Two things I'm *not* at all worried about in Docker land currently:

* The size of the image
* The number of layers

If you combine all the `apt-get` statements as you have in the gist that's fine, as is getting rid of the `upgrade`.

Interesting bits you probably face building the image locally:

* Docker sends the entire directory, including `.git` over for build context so initialization always takes awhile with this repo and `ipython/ipython` unless you did a bare clone. Definitely a won't fix, won't optimize.

I'm open to fig, but for the moment we're using [CoreOS and fleet](https://gist.github.com/rgbkrk/f27c22d0dc414357a1e3). If I set the keys up to using etcd exclusively, I'll be willing to open source deployment setup rather than just that gist. Elasticsearch isn't being used right now until

* we determine what to index and how (including putting all the base64 images in an object store ref'ed by UUID).
* we appropriately handle robots.txt (fun fact: we render whatever people post to us)

bollwyvl:
Great I'll do some more exploring.

I did find adding .git to .dockerignore helped re: initialization

rgbkrk:
> I did find adding .git to .dockerignore helped re: initialization

Oh! Didn't know about that.

bollwyvl:
Added the `.dockerignore`. 

tl;dr: there is a lot more to the docker stuff than what needs to be addressed here. In the nearterm, I think what's on the PR (minimal change to Dockerfile/.dockerignore) is good enough to move forward if it plays well with the existing infrastructure.

re: layer complexity: i was basing it off the rule of thumb from the [best practices](https://docs.docker.com/articles/dockerfile_best-practices/) document: it struck me as reasonable.

Thanks for the links to fleet, etc. That seems like a very capable way to host a service, and I'll investigate that for deployment. Also, it was also good to have the pointers to existing docker layers (nbcachet, etc)... that being said, before I read the I whole fleet thing, I had found that fig handles the environment very nicely: here's a working `fig.yml`
```yaml
nbapp:
  build: .
  links:
  - nbcache
  command: python -m nbviewer --port=8080
  ports:
   - 8080:8080
nbcache:
  image: memcached
```
`fig up` builds and fires up the two images, and it automatically sets the linked-container-based env settings.

Since I didn't know about `jupyter/nbcache`, I found and used the [official repo](https://docs.docker.com/docker-hub/official_repos/) for `memcached`. Is the official repo thing a thing that ipython and or nbviewer should get in on?

I also took a look at `ipython/ipython`'s Dockerfile: perhaps there is merit in basing the nbviewer container off it?

Anyhow: I'll probably not do any more of my own exploring on this PR re: Docker, unless it helps it get the bower stuff merged!

Dockerfile
```diff
@@ -2,26 +2,29 @@
 FROM ubuntu:14.04
 
 MAINTAINER IPython Project <ipython-dev@scipy.org>
- 
+
 # Make sure apt is up to date
 RUN apt-get update
 RUN apt-get upgrade -y
```
rgbkrk: I'll end up removing this line later, but feel free to modify it if I haven't merged yet. I'm running the build right now.

nbviewer/static/bower.json
```diff
@@ -2,10 +2,11 @@
...
   "dependencies": {
-    "Font-Awesome": "~4.0.3",
-    "animate.css": "~3.0.0",
-    "headroom.js": "~0.3.9",
+    "animate.css": "3.1.1",
+    "headroom.js": "0.7.0",
     "requirejs": "~2.1",
-    "moment": "~2.6.0"
+    "moment": "~2.7.0",
+    "bootstrap": "components/bootstrap#~2.3.2",
```
rgbkrk: What's with the pathing here?

rgbkrk:
Ah, now I see an issue with the `.dockerignore` and our current setup:

```
[I 150102 17:35:16 app:144] Not indexing notebooks
[I 150102 17:35:16 app:164] Using in-memory cache
fatal: Not a git repository (or any of the parent directories): .git
[E 150102 17:35:16 app:176] Failed to get git info: Command '['git', 'log', '-1', '--format=%H\n%s\n%cD']' returned non-zero exit status 128
[I 150102 17:35:16 app:249] Listening on port 8080
```

We use that as a fingerprint on the deployed site (look in the footer).

rgbkrk:
For the time being, can you take `.git` out of the `.dockerignore`?

Other than that, everything is perfect and I'm happy to merge this soon.

rgbkrk:
By the way, I just mean that for now we'll take the .git out of the dockerignore and when we change the way the fingerprinting is done we can reintroduce it.

bollwyvl:
Right, right. Thanks for picking the one thing easy to do from my phone!

Changed.

rgbkrk:
Great! Thanks for getting to it so quick. Now to answer your other questions.

rgbkrk:
> I had found that fig handles the environment very nicely: here's a working fig.yml

I'm happy if you want to submit a PR with the fig.yml. I'm using it in [another project](https://github.com/cloudpipe/cloudpipe/blob/master/fig.yml). Before fig was acquired by Docker, I was still letting the orchestration wars play out since I didn't know what was going to last and what had the best trajectory.

> Since I didn't know about jupyter/nbcache, I found and used the official repo for memcached.

That's ok. I was only intending for `jupyter/nbcache` to be the version we run in production, tweaking our default settings. We could honestly crank those up quite a bit, but I haven't yet.

> Is the official repo thing a thing that ipython and or nbviewer should get in on?

Probably not yet. We own the ipython and jupyter namespaces, but we're not at a real release stage for the `ipython/ipython` image. It currently builds straight from master and won't have [tagged builds until IPython 3.x releases](https://github.com/ipython/docker-notebook/issues/32) (as `ipython/ipython:3`, etc.).

> I also took a look at ipython/ipython's Dockerfile: perhaps there is merit in basing the nbviewer container off it?

Possibly after 3.x releases. I'd really like to minimize the dependencies that nbviewer has. The ipython/ipython image is a bit full featured.

We can explore all this in a docker plans issue if you want to open it. I'll be back after lunch to work on some deployment automation.

PR #392, opened 2015-01-10
Build less

### LINKED ISSUE #391: Latex partially rendering in light gray font
I noticed this morning that latex equations are rendering in light gray fonts, whereas they didn't used to. 

It looks bad, because numbers are still rendered in the old style, so equations contain a mix of light gray and black text.

This is not happening in the IPython Notebook (I have version 2.3.0, on Windows and xubuntu 14.04), only on the nbviewer site.

Examples:

http://nbviewer.ipython.org/github/unpingco/Python-for-Signal-Processing/blob/master/Confidence_Intervals.ipynb

http://nbviewer.ipython.org/github/rlabbe/Kalman-and-Bayesian-Filters-in-Python/blob/master/07_Kalman_Filter_Math/Kalman_Filter_Math.ipynb

bollwyvl:
This adds LESS to build, combine and compress the CSS assets. It fixes #391, and reduces the number of  requests by a fair amont:
screenshot from 2015-01-10 18 14 00

There is a new invoke task: `invoke less`. It can generate source maps.

This combines all of the CSS from `layout.html` into `static/build/styles.css` and all of the CSS from `notebook.html` into `static/build/notebook.css`.

Wherever possible, the existing CSS is being interpreted as LESS, so it can be compressed when served.

The Dockerfile now uses the invoke tasks.

### DIFF
--- a/nbviewer/templates/layout.html
+++ b/nbviewer/templates/layout.html
@@ -54,12 +54,7 @@
     <!--NREND-->
 
     <!-- Le styles -->
-    <link href="/static/css/docs.css" rel="stylesheet">
-    <link href="/static/components/bootstrap/css/bootstrap.min.css" rel="stylesheet">
-    <link href="/static/components/bootstrap/css/bootstrap-responsive.css" rel="stylesheet">
-    <link href="/static/components/animate.css/animate.min.css" rel="stylesheet">
-    <link href="/static/components/font-awesome/css/font-awesome.min.css" rel="stylesheet">
-    <link href="/static/css/nbviewer.css" rel="stylesheet">
+    <link href="/static/build/styles.css" rel="stylesheet">
 
 
     <!-- Le HTML5 shim, for IE6-8 support of HTML5 elements -->
--- a/Dockerfile
+++ b/Dockerfile
@@ -24,7 +24,7 @@ RUN apt-get install -y -q python python-dev python-pip
 RUN apt-get install -y -q libzmq3-dev sqlite3 libsqlite3-dev pandoc libevent-dev libcurl4-openssl-dev libmemcached-dev nodejs nodejs-legacy npm
 
 # asset toolchain
-RUN npm install -g bower
+RUN npm install -g bower less
 
 # install IPython 2.x branch
 WORKDIR /srv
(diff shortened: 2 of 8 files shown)


bollwyvl:
Tests are passing locally: looks like some API rate limiting, etc.

nbviewer/templates/layout.html
```diff
@@ -54,12 +54,7 @@
     <!--NREND-->
 
     <!-- Le styles -->
-    <link href="/static/css/docs.css" rel="stylesheet">
-    <link href="/static/components/bootstrap/css/bootstrap.min.css" rel="stylesheet">
-    <link href="/static/components/bootstrap/css/bootstrap-responsive.css" rel="stylesheet">
-    <link href="/static/components/animate.css/animate.min.css" rel="stylesheet">
-    <link href="/static/components/font-awesome/css/font-awesome.min.css" rel="stylesheet">
-    <link href="/static/css/nbviewer.css" rel="stylesheet">
+    <link href="/static/build/styles.css" rel="stylesheet">
```
rgbkrk: Awesome.

rgbkrk:
If/when we ship this I need to make sure to clear the cache on Fastly.

rgbkrk:
Deploy will be underway just as soon as [Docker Hub finishes the build](https://registry.hub.docker.com/u/jupyter/nbviewer/builds_history/87496/).

rgbkrk:
We're live!

screenshot 2015-01-11 20 16 57

bollwyvl:
Great!

There are some follow-on LESS possibilities, mostly related to bootstrap, that could be tackled now:
- LESS-ify existing nbviewer.css
  - cherrypick doc.css instead of bringing the whole thing
- go to 3? (was discussed on gitter, i think)
  - could also see waiting until ipython 3.0 lands
    - is the jupyter branding available as a LESS-ful theme?
- de-bootstrap html, add semantic classes and re-bootstrap with LESS
  - would make going to 3 easier (potentially LESS only)
  - may not be worth it if there are significant design changes
- restore old customizations for collapse threshhold
  - would require changing bower version to "real" 2.3.2 (components/bower only has built assets)
    - would also give us back collapse.js, which is the only part of bootstrap.js that is used

happy to help in any way i can! i'll try to make the meeting on tuesday, but my schedule may not be my own :)

PR #393, opened 2015-01-12
Frontpage updates

rgbkrk:
Noticed that @camdavidsonpilon had changed the URL of the initial chapter, so I updated it here.

In addition, I moved a newer notebook up and took out some older notebooks from 2012.

@twiecki, I cut a gist of yours out that had deprecated bits from Pandas. Would you be happy with swapping to https://github.com/twiecki/financial-analysis-python-tutorial or would you rather submit a different notebook in its place?

### DIFF
--- a/nbviewer/frontpage.json
+++ b/nbviewer/frontpage.json
@@ -34,7 +34,7 @@
       },
       {
         "text": "Probabilistic Programming",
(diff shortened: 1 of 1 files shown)


twiecki:
Thanks @rgbkrk. This is the most up-to-date tutorial: http://nbviewer.ipython.org/github/quantopian/zipline/blob/master/docs/tutorial.ipynb

Issue #409, opened 2015-02-13
nbviewer doesn't accept branchnames with slashes


remram44:
On my branch "testbranch": https://github.com/remram44/ipython-nbviewer/blob/testbranch/test.ipynb
Renders correctly: http://nbviewer.ipython.org/github/remram44/ipython-nbviewer/blob/testbranch/test.ipynb

On my branch "wrong/branch": https://github.com/remram44/ipython-nbviewer/blob/wrong/branch/test.ipynb
Error 404: http://nbviewer.ipython.org/github/remram44/ipython-nbviewer/blob/wrong/branch/test.ipynb

minrk:
I think we need to escape branch names

remram44:
I don't think there are characters that are valid in branch names but not in URLs. The exception here is the slash `/`, which is technically valid in URLs but that you use as separator.

If you can't get your parser to handle these, you could replace slashes in branch names with one of the invalid characters like the colon `:`.

Note that non-ascii characters are typically accepted in branch names, but percent-encoding should be enough for these.

bollwyvl:
Unfortunately, I think anything that gets us away from being able to accept a copy/pasted github URL is a bad idea... but perhaps we can gracefully handle it with some redirects. It seems like everyone that interacts with github [eventually](https://github.com/travis-ci/travis-api/pull/146) [runs](https://github.com/npm/npm/commit/ae1d590bdfc2476a4ed446e760fea88686e3ae05) up against this.

One thought: For the tree view, we already pull branches/tags. Perhaps, on a 404, we pull the `refs`, and figure out if `ref + / + path` starts with a `ref` that isn't the naive one we already tried. Then escape, and redirect. Luckily, this won't be a problem for commits (they will never have slashes).

remram44:
> Unfortunately, I think anything that gets us away from being able to accept a copy/pasted github URL is a bad idea...

Do people actually construct the URL from their Github URL? Or do you mean pasting into the "gistnorurl" input on the index? (because this one posts to `/create/`, which redirects).

Carreau:
I do navigate github and re-write urls. There are aven bookmarklets to do that.

remram44:
Because nbviewer has no way of knowing where the ref ends, it could just try them in turns:
`/github/remram44/ipython-nbviewer/blob/wrong/branch/name/here/test.ipynb`:
* ref: `wrong` path: `branch/name/here/test.ipynb
[...]

PR #411, opened 2015-02-19
Accept GitHub branch/tags with slashes

bollwyvl:
This proposes one approach to accepting refs that contain slashes, as in #409. Github accepts these on their public URL scheme, but not on their API, which actually wants SHAs... though it does support "normal", non-slashed refs, which has created this whole expectation.

This approach pulls down the tags and branches, and checks them for the first matching name, then replaces the `ref` with the SHA of the first named ref. Because of this, we will likely want to revisit caching the upstream API calls, as now any 404 is much more expensive.

While it's not possible for one slashed ref name to be contained in another (because they map to files in `.git`), it probably WOULD be possible for some combination of ref + path...
- this
  - ref `foo/bar`
  - path `baz`
- that
  - ref `foo`
  - path `bar/baz`

... but now you're just being evil.

There is still some work to do, as it changes the refs in-line to minimize total changes. The tree view, for example, now shows the SHA instead of the symbolic name, which is ugly. Wanted to get this out there in the meantime.

### DIFF
--- a/nbviewer/providers/github/handlers.py
+++ b/nbviewer/providers/github/handlers.py
@@ -47,6 +47,21 @@ class GithubClientMixin(object):
         return self._github_client
 
 
+class RefMixin(object):
+    @gen.coroutine
+    def refs(self, user, repo):
+        """get branches and tags for this user/repo"""
+        ref_types = ("branches", "tags")
+        ref_data = [None, None]
+
+        for i, ref_type in enumerate(ref_types):
+            with self.catch_client_error():
+                response = yield getattr(self.github_client, "get_%s" % ref_type)(user, repo)
+            ref_data[i] = json.loads(response_text(response))
+
+        raise gen.Return(ref_data)
+
+
 class RawGitHubURLHandler(BaseHandler):
     """redirect old /urls/raw.github urls to /github/ API urls"""
     def get(self, user, repo, path):
@@ -93,7 +108,7 @@ class GitHubUserHandler(GithubClientMixin, BaseHandler):
             ))
         provider_url = u"https://github.com/{user}".format(user=user)
         html = self.render_template("userview.html",
-            entries=entries, provider_url=provider_url, 
+            entries=entries, provider_url=provider_url,
             next_url=next_url, prev_url=prev_url,
             **PROVIDER_CTX
         )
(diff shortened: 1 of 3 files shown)


bollwyvl:
Fixed up the symbolic names. Getting a few spurious test fails, but think this pretty much answers the mail. There are still ways to get to the SHA-based links, but then, that happens fairly often on GH anyway.

Carreau:
This sleeps through the cracks, do you want to attempt a rebase ?

bollwyvl:
Perhaps we wait until #443 lands, and then bring this back in?

Carreau:
Sure you are mostly the boss with Kyle on nbviewer.

nbviewer/providers/github/handlers.py
```diff
@@ -115,12 +130,34 @@ def get(self, user, repo, ref, path):
...
             return
         path = path.rstrip('/')
+        symbolic_ref = None
+
         with self.catch_client_error():
-            response = yield self.github_client.get_contents(user, repo, path, ref=ref)
+            branches, tags = yield self.refs(user, repo)
 
-        contents = json.loads(response_text(response))
+            try:
```
mrterry: this is getting really complicated.  this will be a beast to test.

mrterry:
Nothing in the github api prevents you from pulling data from weirdly named branches.  The problem is disambiguating which part of `/github/user/repo/blob/maybe/branch/maybe/path` corresponds to the branch and which to the path.

    import requests                                                                 
    from urllib.parse import quote                                                  

    # resource is as weird as possible.  non root directory, unicode filename, branch with slashes                                                                             
    url = quote('/repos/mrterry/wonky/contents/directory/' + 'שבור')                
    params = {'ref': 'slash/branch'}                                                
    resp = requests.get('https://api.github.com' + url, params=params)

An alternate implementation would be to:

1. Extract user & repo from url.  This part is non-ambiguous.
1. Ask for the branches for the repo.  `/repos/username/reponame/branches`
1. Split the two and use the existing machinery to render.

I'm not sure what the authoritative url should be for this case.  Maybe just leave it `/mabye/branch/maybe/path` and walk away?

mrterry:
example of fetching through the API.  takes 3 requests unfortunately.

https://gist.github.com/mrterry/0aedd638eb273810babc

Carreau:
@bollwyvl I'm closing to keep the global Jupyter-org PR queue under control, but feel free to reopen.

Thanks !

PR #390, opened 2015-01-10
adding format /slides/ for reveal.js

bollwyvl:
Reveal.js `master` just went [3.0](https://github.com/hakimel/reveal.js/releases/tag/3.0.0): congratulations, @hakimel, and thanks for all your hard work! :trophy:

This, and the fact that bower has quieted down (#378 and #388), made it an auspicious time to reintroduce the reveal.js-based /slides/ format originally thieved from #150, and to finish off the review debt of #359 incurred by my cowboy coding. While I couldn't help myself, and did move formats out of `app`, this should be a much more manageable chunk of changes to review.

From a UI point of view, it looks exactly the same. I certainly welcome feedback on any choices made there!

Under the hood, differences from the last slides PR:
- moved formats into their own file, as it was starting to get unwieldy
  - these still need to probably refactored (in another PR...) into either:
    - singleton classes
    - a deep config object
- passing the whole format to render (API change!) so that the reveal-specific post-processing can be done by the format

Thanks for all the support in getting (back) here!

PR #412, opened 2015-02-28
Use IPython 3.0

bollwyvl:
Hooray IPython 3.0 :cake:! Welcome to the future :monorail:!

This upgrades nbviewer's IPython requirement to `IPython[notebook]>=3.0`.

Since all good things come in threes, I'd love to see (IPython, Bootstrap & Reveal (ahem, #390)) 3.x all wrapped up as a 0.3.0 release!

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -28,7 +28,7 @@ RUN npm install -g bower less
 
 # install IPython 2.x branch
 WORKDIR /srv
-RUN git clone --depth 1 -b 2.x https://github.com/ipython/ipython.git
+RUN git clone --depth 1 -b 3.x https://github.com/ipython/ipython.git
 WORKDIR /srv/ipython
 RUN git submodule init && git submodule update
 RUN pip install .
--- a/Dockerfile
+++ b/Dockerfile
@@ -35,15 +35,19 @@ RUN pip install .
 
 RUN pip install invoke
 
-ADD . /srv/nbviewer/
+ADD ./requirements.txt /srv/nbviewer/
 WORKDIR /srv/nbviewer
-
-RUN invoke bower
-RUN invoke less
 
 RUN pip install -r requirements.txt
 
+ADD ./tasks.py /srv/nbviewer/
+ADD ["/nbviewer/static/bower.json", "/nbviewer/static/.bowerrc", "/srv/nbviewer/nbviewer/static/"]
+RUN invoke bower
+
 EXPOSE 8080
+
+ADD . /srv/nbviewer/
+RUN invoke less
 
 # To change the number of threads use
 # docker run -d -e NBVIEWER_THREADS=4 -p 80:8080 nbviewer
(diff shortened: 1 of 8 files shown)


bollwyvl:
Error appears to be Travis->Github DNS...

bollwyvl:
Hooray, build fixed. Docker build improved somewhat as well.... mainly to prevent re-pip/bower-ing after every file change for local test... though this could be taken out, if that's too far outside the scope.

bollwyvl:
Gah, or the build is not fixed. Again, seeing lots fo `40x` from github. My oh my.

Dockerfile
```diff
@@ -28,23 +28,27 @@ RUN npm install -g bower less
...
+ADD ./requirements.txt /srv/nbviewer/
 WORKDIR /srv/nbviewer
 
-RUN invoke bower
-RUN invoke less
-
 RUN pip install -r requirements.txt
 
+ADD ./tasks.py /srv/nbviewer/
+ADD ["/nbviewer/static/bower.json", "/nbviewer/static/.bowerrc", "/srv/nbviewer/nbviewer/static/"]
```
rgbkrk: Well this is new to me. How does `ADD` work with a list?

README.md
```diff
@@ -63,6 +63,18 @@ docker build -t nbviewer .
...
+
+The Notebook Viewer uses `memcache` in production. To locally try out this
+setup, a [docker-compose](https://docs.docker.com/compose/) configuration is
+provided to easily start/stop the `nbviewer` and `memcache` containers together:
+
+#### Run
+
+```shell
+docker-compose up
+```
```
rgbkrk: Thank you for doing this! I've been waiting on compose to release, had not realize it had shipped.

requirements.txt
```diff
@@ -1,4 +1,3 @@
-tornado>=3.1.1
```
rgbkrk: Technically this makes sure our underlying server matches tornado. I couldn't imagine IPython notebook changing (like it did from twisted) again, but still.

rgbkrk:
Looking good. If the docker build is going well, I say ship this. The only comment I made that I care about a response is the `ADD` with a list.

Dockerfile
rgbkrk: Ah, it's intended to be `["<src>"... "<dest>"]`. Should these be `./nbviewer/static/...`?

bollwyvl: You're right! Will fix!

README.md
bollwyvl: Yeah they were very sneaky. It was confusing for a while as they still were using the fig gh.

requirements.txt
bollwyvl: True. Will fix!

bollwyvl:
Addressed above comments. Changed the links to bower stuff... was definitely a typo, though ADD will [only, ahem, add](http://stackoverflow.com/questions/27068596/how-to-include-files-outside-of-dockers-build-context) things inside the current build context/dir. But it looked weird, for sure, and the `./` is easier to understand!

Eyeing the shiny green button, but wouldn't mind some more :eyes:! In the meantime will start working up the bootstrap3 stuff!

rgbkrk:
Looks great to me. Thanks @bollwyvl!

PR #413, opened 2015-03-02
Use Bootstrap 3.x

bollwyvl:
This uses Bootstrap 3.x, with more LESS. 

With more LESS, we need autoprefixing, and want compression, which has lead me to introducing a `package.json`... slippery slope indeed!

Home page
- nothing too drastic.

Mobile Home page
- a little bit of love

A notebook
- bringing in the "notebook shadow"
- the server icon up top indicates the current kernel, if available


minrk:
I would love to pull bower out of the repo and make running bower and less a required part of deployment.

rgbkrk:
Loving this so far, especially the now working front page on mobile!

I'm likely not going to be a fan of the document shadow since it should just flow as rendered HTML.

bollwyvl:
Yeah figured before we went down search/gallery road we should get the
baseline home better. What would other ideal mobile (non)features be?
Haven't really addressed anywhere else, though i think generally 3.x is at
least as good as 2.x.

I am OK with removing the shadow... I was just trying to bring elements
from the 3.0 interface in, where it made sense. Corollary: Is there a
kernel icon?

bollwyvl:
@minrk upstream? 

For nbviewer, I still had the idea in mind that we eventually cut a pip release of nbviewer at some point with all the built assets bundled.

In related wishes, I'd love a sudoless install to be able to do fast travis builds. Cache sounds nice.

One more *3 question... Any reason not to think about supporting py3? Easy enough to add to the test regime...

Carreau:
agreed sudo-less woudl be neat, and love the new page. I would even cut the rounded corner even more.

bollwyvl:
Removed more corners and the notebook dropshadow.

Some more mobile tweaks: `img-responsive` for sub-768px screens.

Additionally, as I got tired of making screenshots, added a casper script generator... might just be tired, as it seems to not really do the mobile viewport sometimes, but it's a start:
http://bit.ly/1M2BFid

rgbkrk:
> One more *3 question... Any reason not to think about supporting py3? Easy enough to add to the test regime...

I'd love it if we were just 100% Python 3.

rgbkrk:
Let's ship it and see how it goes.

rgbkrk:
Ruh roh.

screenshot 2015-03-03 10 59 48

Address for this node is http://104.130.161.114/

rgbkrk:
You know what, it's fine on other pages. I think this is an artifact of old memcache with new markup.

rgbkrk:
Yeah, we're totally good. Thanks @bollwyvl!

rgbkrk:
The browser cache will likely collide for some users on nbviewer.ipython.org. It did for me. :wink:

bollwyvl:
Hooray! I'll start the cascade of closing the tickets i referenced from here!

PR #421, opened 2015-03-07
support python3

bollwyvl:
This should add support for python 3.

The changes are minimal to get it to run: just handling itervalues(2)/values(3), which only occurs once.

This also updates travis and docker.

The initial [fail on 2.7 and 3.4](https://travis-ci.org/bollwyvl/nbviewer/builds/53482911) are both related to unicode: surprise! I've fixed the 2.7 issue in my hacked test base, but 3.4 is being tricky.

Will continue to poke on this, but welcome any insights!


nbviewer/handlers.py
```diff
@@ -693,8 +693,12 @@ def get(self, user, gist_id, filename=''):
             entries = []
             ipynbs = []
             others = []
+            
+            iter_files = (
+                getattr(files, "itervalues", None) or getattr(files, "values")
+            )
 
-            for file in files.itervalues():
+            for file in iter_files():
```
minrk: My Python 3-centrism says to use `files.values()` in places like this, and let Python 2 be slightly less efficient than Python 3. These aren't large enough containers for the optimization to ever be critical.

bollwyvl:
Hooray! All fixed.

I'll change it to values, and nobody will be the wiser.

rgbkrk:
> I'll change it to values, and nobody will be the wiser.

:wink:

bollwyvl:
Ah, good, didn't error here from github api... original push did.

I've also added 3.3, since that's what ships on several distros in the wild...

Interested in feedback on Dockerfile... it's some fairly drastic changes informed by other things I have been learning which might well break peoples' downstream images... if there are any, other than mine :)

Also, humorously, docker-compose doesn't work with python3 yet... except they already use `six`. Weird.

bollwyvl:
Needs rebase with #430.

bollwyvl:
Running the docker-compose setup (which includes memcached) revealed that the cache handling wasn't ready for python3... and in fact is not tested at all. Not sure if that is a blocker for this, as I was able to make the fix in d94f813, but should be considered.

The test fail is a random `400` from the github api.

Carreau:
Reran failing tests on Travis.
Passing now.
+1

bollwyvl:
Cool, thanks for the review!

I guess the question remains whether we actually bump up the Docker image
to use python3 by default (and therefore in production).

minrk:
I think using Python 3 by default is a good plan.

rgbkrk:
If you want to switch to using `python3` in the `CMD` line, that's fine by me!

bollwyvl:
Despite the `400` test fails, it looks good to me! I guess someone who can trigger the build should press the button so we can manage the deployment, and not have any surprises :)

Carreau:
@bollwyvl you should be able to restart the failing build when it's random travis faillure. 

I did it for you.

bollwyvl:
Hm, i'll have to look into that. The UI changed not too long ago, still getting used to it.

Carreau:
capture d ecran 2015-04-20 a 12 08 18

this circular arrow on the right of either all jobs (rerun all the things), specific job (rerun this job). 

Travis CLI tool allow that too I guess, but I don't know how,

bollwyvl:
Cool, thanks!

rgbkrk:
Now that #443 is in, care to rebase and we'll carry on with Python 3 for the sake of the future?

bollwyvl:
yes indeed!

bollwyvl:
That's the worst commit history I've done since using git. I would not be averse to a squash merge, force push, but I am keeping the history for a sec in case I did something really wrong. open to suggestions o how this might have been done better!

Also, we really have to do something about the tests re: Github API fails! That travis refresh button has gotten to be like a slot machine for me.

Carreau:
Would github API fails be a rate limited thing ? Could we encrypt a token
to have less limitations ?

You can also squash in a new branch and open another PR if this make you
feel bad.

bollwyvl:
I don't feel bad about it, it just took a long time!

there are keys set up in there, I assumed the merge request builds were
using them: i don't pay attention to my builds much. i wonder: does
production nbviewer use the same API key as the encrypted ones on travis?
Perhaps that's part of our problem.

rgbkrk:
As I said before about tests, we need to be mocking instead of hitting the full server.

> Would github API fails be a rate limited thing ? Could we encrypt a token
to have less limitations ?

Yes, we get rate limited from Travis frequently. Encrypted Travis tokens only work on master. We can run those for integration tests but not PRs.

bollwyvl:
Ah, my misunderstanding.

Well, we should look hard at that. Is there a good tool out there for
moving the github API for a bunch of static json? We will want to emulate
errors, etc. I guess either a node or python tool would be best.
nose-github-api?

rgbkrk:
github3.py is the most featureful and has mocks as part of their testing setup. We should mirror that as much as possible.

minrk:
Going through stale PRs. It looks like this one should be ready to merge. Any objection?

rgbkrk:
No objections.

Issue #424, opened 2015-03-13
Running docker container fails


mpelko:
Following:

    $ docker pull jupyter/nbviewer
    $ docker run -p 8080:8080 jupyter/nbviewer

I get:

    Traceback (most recent call last):
    File "/usr/lib/python2.7/runpy.py", line 162, in _run_module_as_main
      "__main__", fname, loader, pkg_name)
    File "/usr/lib/python2.7/runpy.py", line 72, in _run_code
      exec code in run_globals
    File "/srv/nbviewer/nbviewer/__main__.py", line 2, in <module>
      main()
    File "nbviewer/app.py", line 175, in main
      frontpage_sections = json.load(f)
    File "/usr/lib/python2.7/json/__init__.py", line 290, in load
      **kw)
    File "/usr/lib/python2.7/json/__init__.py", line 338, in loads
      return _default_decoder.decode(s)
    File "/usr/lib/python2.7/json/decoder.py", line 366, in decode
      obj, end = self.raw_decode(s, idx=_w(s, 0).end())
    File "/usr/lib/python2.7/json/decoder.py", line 382, in raw_decode
      obj, end = self.scan_once(s, idx)
      ValueError: Expecting ',' delimiter: line 50 column 7 (char 1566)

bollwyvl:
Thanks, will definitely look into this!

Presumably, the frontpage json is not valid, and we should probably test
for that.

dennisobrien:
The same issue also occurs when running locally.  I manually patched my local version since there is already a PR for this.

I like the idea of tests for well-formed json.  Nice work @bollwyvl

bollwyvl:
haven't merged the tests, but this should be good now!

PR #426, opened 2015-03-14
Test json files

bollwyvl:
This adds tests for well-formedness of JSON files that are part of the project, and fixes the invalid `frontpage.json` from #424. Instead of just parsing every JSON file in the project (think `node_modules`), I've just ticked them off explicitly.

`package.json` and `bower.json` are basically impossible to validate, but since we already have `jsonschema` around from IPython, I have added a schema for the front page. This should give us some good assurance of avoiding surprises in the future. Or something.


Carreau:
+1 ! Thanks a lot !

PR #430, opened 2015-03-20
Docker: FROM ipython/ipython:3.x

bollwyvl:
Deployment is getting a little behind IPython 3.x master. With the big split is coming, this might be the best way to have an image that is close to what is expected by upstream development.

Also some little movements of things that don't change much and were pretty late in the build.

And whitespace. Always whitespace.

@rgbkrk what were your reasons for not wanting to do this earlier?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,34 +1,15 @@
 # Using the Ubuntu image
-FROM ubuntu:14.04
+FROM ipython/ipython:3.x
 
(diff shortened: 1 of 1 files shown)


rgbkrk:
> what were your reasons for not wanting to do this earlier?

That it didn't exist yet. Now it does!

It also has more dependencies than we actually need. I'm fine with doing this though.

Issue #431, opened 2015-03-20
Unicode notebook rendering pretty bad


bollwyvl:
Notebooks with unicode are rendering poorly. This doesn't occur with straight `nbconvert`, so must be on this side.

Cursory exploration indicates that just making [this line](https://github.com/jupyter/nbviewer/blob/master/nbviewer/handlers.py#L530):

```python
        html = self.render_template(
            "formats/%s.html" % format,
            body=u'{}'.format(nbhtml),
```

will fix the issue.

>Hi Dami\u00e1n Avila ;-)

>That is really cool, but I noticed that you used unicode representations to get a properly rendered version of your HTML.  Did you do that by hand or do you have an automated way of substituting the accents for unicode code?

>Without that the non-English slides look awful:

>http://nbviewer.ipython.org/format/slides/gist/ocefpaf/cf023a8db7097bd9fe92

>Thanks,

>-Filipe

bollwyvl:
I also have the sneaking suspicion that going to python3 (#421) would also fix this... but it's probably worth being explicit as long as we are supporting py27.

bollwyvl:
So a resaved minimal subset of that notebook, [reuploaded to gist](https://gist.github.com/bollwyvl/3c50eaae790db5bde0a8) and served [does not demonstrate the condition](http://nbviewer.ipython.org/format/slides/gist/bollwyvl/3c50eaae790db5bde0a8#/), though when viewed in the raw, does show the encoding creepypasta.

Additionally, a [pass-through to rawgit](https://cdn.rawgit.com/ocefpaf/cf023a8db7097bd9fe92/raw/49ec93a0a688c8aaad8f9f4c0c830471a2dfe9a0/OMP.ipynb) also [does not demonstrate the condition](http://nbviewer.ipython.org/urls/cdn.rawgit.com/ocefpaf/cf023a8db7097bd9fe92/raw/49ec93a0a688c8aaad8f9f4c0c830471a2dfe9a0/OMP.ipynb).

A checked out version of the original file served through `/localfile/` does not demonstrate the condition. Still trying to create a minimal test case.

bollwyvl:
The shortened example isn't triggering the condition because then the content is available in the already-correctly typed JSON response. If the result is truncated due to size, then the raw_url is used... and this is in turn served as `text/plain` with no encoding. Changing that appears to fix it. PR incoming....

bollwyvl:
Should be fixed on next deployment.

PR #432, opened 2015-03-21
Force UTF-8 on notebooks from truncated gists

bollwyvl:
Turns out #431 was based on a gist being too big, getting truncated, getting fetched, and then not being properly decoded. This adds a test and a fix, by adding an option encoding argument to `response_content` which ignores the header.


Carreau:
Thanks !

Issue #428, opened 2015-03-18
Refactor github provider as template for other providers


bollwyvl:
Right now, the github/gist provider is:
- ~600 lines of handlers.py
- the github.py async client

Refactoring this into some common patterns that can be reused for other providers would be the first step towards #371 (gitlab), #302 (webgit), and #402 (stash/bitbucket) and #427 (google drive). Additional providers:
- allura/sourceforge
- dropbox
- confluence

## `Provider`
The theoretical base class, `Provider`, might like to have implementations include:
- `handlers`: a la notebookapp, this is all `handlers.py` would really care about, and would handle the format story (a whole separate thing!)
- `entity(entity)`: user/org/project json-like for templates
- `tree(entity, collection, ref=None)`: repo json-like for templates
- `file(entity, collection, file)`: would a notebook need something special? The urls will be poor predictors for the general case.
- `refs(entity, collection, file=None)`: history, etc. presumably even drive would support this

Not all of these would be necessary for a 1.0 of a given implementation.

## `Client`
Not sure if this is just an implementer convenience thing, but consistent logging, error reporting, caching, etc. would be handy, even if the individual methods were completely different from implementation to implementation... only the *Provider would care.

### Monolith
Perhaps the right file structure, if everything is to live in nbviewer itself is:
```
- app.py
- handlers.py
- ...
- providers/
  - provider.py
  - client.py
  - github/
    - provider.py
    - client.py
  - stash...
  - drive...
  - gitlab...
  - gitweb...
  - allura...
  - dropbox...
  - confluence...
```

And turn the providers on and off with the Configuration.... which may be a prerequisite for doing this correctly. 

### Plugins
Another option would be to just go ahead and use distribute/setuptools entry points, and break the providers into separate repos
```python
setup(
    # ...
    entry_points = {'nbviewer.provider': 'github = nbviewer_github:GithubProvider'}
)
```

And manage them, for example, in your docker build....
```dockerfile
FROM jupyter/nbviewer
RUN pip install nbviewer-github
CMD ["python", "-m", "nbviewer"]
```

Personally, I favor this approach.

Let's chart a course to the future of publishing :rocket: !

bollwyvl:
Looking at this more, the basic url handler could also benefit from being able to render a directory index view such as apache, WebDAV or svn might provide. This might be a better place to set the pattern than a synthetic provider, as it has a client, etc.

[...]

Issue #405, opened 2015-02-04
Build search functionality 


danabauer:
This is a high-level issue to add search functionality to nbviewer. Some of the steps are:
- [ ] detect robots.txt from the originating URL so we can regard some as private
- [ ] determine how to index content in elasticsearch
- [ ] figure out how to strip out base64 encoded images so they are not indexed
- [ ] slurp in all markdown and code cells
- [ ] expose an api for the search box proposed in #252 

Feedback @rgbkrk?

rgbkrk:
Looking great! I just updated the list into a checklist and we can flesh it out a bit more.

PR #442, opened 2015-04-15
Parse robots.txt from originating URLs

rgbkrk:
This totally doesn't seem to be working on http://nbviewer.ipython.org/url/jdj.mit.edu/~stevenj/IJulia%20Preview.ipynb

:-/

Posting this early to help with PyCon sprinters that want to touch the same stuff.

### DIFF
--- a/nbviewer/handlers.py
+++ b/nbviewer/handlers.py
@@ -592,7 +592,7 @@ class URLHandler(RenderingHandler):
 
         rfp = robotparser.RobotFileParser()
         rfp.set_url(robots_url)
-        rfp.parse(robotstxt)
+        rfp.parse(robotstxt.splitlines())
         public = rfp.can_fetch('*', remote_url)
 
         response = yield self.fetch(remote_url)
--- a/nbviewer/handlers.py
+++ b/nbviewer/handlers.py
@@ -22,9 +22,11 @@ try:
     # py3
     from http.client import responses
     from urllib.parse import urlparse
+    from urllib import robotparser
 except ImportError:
     from httplib import responses
     from urlparse import urlparse
+    import robotparser
 
 from tornado import web, gen, httpclient
 from tornado.escape import utf8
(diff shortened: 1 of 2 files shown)


rgbkrk:
The robots.txt parser seems to work right when I was directly using it with `read()`:

```
>>> from urllib import robotparser
>>> rp = robotparser.RobotFileParser()
>>> rp.set_url("http://jdj.mit.edu/robots.txt")
>>> rp.read()
>>> rp.can_fetch("*", "http://jdj.mit.edu/~stevenj/IJulia%20Preview.ipynb")
False
>>> rp.can_fetch("*", "http://jdj.mit.edu/~stevenj/IJulia%20Preview.ipynb")
True
```

rgbkrk:
Ah, parse doesn't take the whole file. It takes a list of lines.

```
rp = robotparser.RobotFileParser()
rp.set_url("http://test.com")
rp.parse('User-Agent: *\nDisallow: /'.splitlines())
rp.can_fetch('hello', "/")
```

rgbkrk:
With changes from @melignus we're not passing on whether a notebook is indexable or not. This will help for #405.

bollwyvl:
Awesome stuff.

As we think about the url/s handlers as the baseline for other providers (#428), what can we learn from this work? Is a secret gist also not public? Google docs links?

Also, based on this, can we tell whether it is okay to scrape the page and look for ipynbs and directories?

bollwyvl:
Also, does this need any form of configuration?

melignus:
No configuration. As it stands now it's only adding a curtesy noindex,nofollow in the viewer if the same settings are set in the robots.txt to honor the origins settings.

I would think if the robots.txt specifically states nofollow,noindex, it would be polite not to scrape the page for more notebooks not in the specific path?

rgbkrk:
A secret gist is also not public. This one only handled the URL case (not a special GitHub provider). I haven't looked into Google docs. 

In general, we're not currently scraping the page to look for other notebooks and directories. That makes me the most comfortable operationally and for the sake of development resources.

> Also, does this need any form of configuration?

If it comes up later for alternative deployments of nbviewer, we can address it. I'm only worried about nbviewer.ipython.org being a fair citizen server of the interwebs for the moment. Propagating noindex (and in turn our own robots.txt) seemed like a decent start.

It might have been better if we had made the default `noindex` instead of both `noindex` and `nofollow` though.

PR #447, opened 2015-04-29
[wip] Use setuptools entry_points for providers

bollwyvl:
This builds on #443, replacing the homegrown extension mechanism cargo culted from IPython with `setuptools` `entry_points`.

Basically, each feature is an entry point: `dropbox`, for example, only implements `uri_rewrite`, while `github` does both `uri_rewrite` and `default_handlers`. The ordered list of config features is re-implemented with a weight scheme (not the most elegant, open to suggestions).

Here is a [strawman implementation](/bollwyvl/nbviewer-gitlab) of a gitlab provider (the URLs will take some work). I like that to use it, an end user need only install it, and it takes care of the rest.


minrk:
That sounds like the presence of a provider would make it enabled by default on startup without request. That doesn't seem desirable.

bollwyvl:
Yes, that is the interpretation of the setuptools entry_point approach I have taken. Aside from the kludgy ordering, I think it makes a lot of sense for the direction we want to go with nbviewer towards hostability. If I am using, say, gitlab, I want to just `pip install nbviewer nbviewer-gitlab` and fire it up with some environment variables for API keys. If it's not needed, it shouldn't be installed.

Of course, we could add the configuration construct back in: nose, for instance, uses entry_points but doesn't autoload plugins, i.e. `--with-coverage`, etc. So the case above could become `python -m nbviewer --with-gitlab`, which is not too onerous.

minrk:
I would be more comfortable with requiring explicit request. Every example I can think of (nose, pygments, etc.) requires users to request the use of any plugins.

minrk:
It would also mean that for testing, I would need to keep installing and uninstalling plugins, rather than changing startup args or config.

bollwyvl:
Ok, how about this: go with the `--with-<provider>` and `--without-<provider>`, where <provider> is the name in the entry point. Everything already installed gets grandfathered in, so the default is actually:
```bash
python -m nbviewer --with-github --with-gist --with-url --with-dropbox --with-url
```
With order being unimportant.

For my example above, if I wanted to just have an nbviewer that only saw my firewalled gitlab, it would be like this:
```shell
GITLAB_API_TOKEN=FoQG4X5byjPZjFbmLdKo \
GITLAB_API_URL=http://example.com \
    python -m nbviewer \
      --with-gitlab \
      --without-github \
      --without-gist \
      --without-github \
      --without-url
```
Open to other suggestions.

I'd like to see the plugin infrastructure get right: there was talk on [gitter about analytics](https://gitter.im/jupyter/nbviewer?at=553fffa42b405ca55d5ad24a), for example so we will probably see this pattern again. The production machine would run this (getting the key out of the template!)
```
GA_API_KEY=UA-38683231-2  python -m nbviewer --with-ga
```

While my fictional company above could use pivik:
```
PIVIK_API_URL=http://pivik.example.private python -m nbviewer --with-pivik
```

minrk:
That proposal sounds pretty good to me. As a minor note, GA might be a case I wouldn't bother doing as a plugin, I would probably enable it iff GA_API_KEY is defined.

bollwyvl:
Yeah, I can see the benefit there. Perhaps tornado's command thing can do casting aliases...

bollwyvl:
I've added the `with_` and `without_` flags... tornado likes underscores, not hyphens, though it will answer to hyphens on the command line.  Seems okay. 

Travis tells me that I am doing something wrong with the rewrites: I'll check it out. Speaking of tests: I guess to test `with_` I'll have to make a fake module and install it :)

`nose` also allows its plugins to define options for the command line, not just environment variables. I wonder if that is a thing that we should also do... that wouldn't really qualify as a feature in my current design... but then maybe that's an argument for just having one entry point be a wrapper class that can then implement a `define_options`, in addition to `handlers` and `uri_rewrites`.

Anyhow, below is what `--help` looks like now (I've gone and added groups to everything to pretty it up a bit).

```
Usage: __main__.py [OPTIONS]

Options:

  --help                           show this help information

/home/weg/.virtualenvs/nbviewer/lib/python2.7/site-packages/tornado/log.py options:

  --log_file_max_size              max size of log files before rollover
                                   (default 100000000)
  --log_file_num_backups           number of log files to keep (default 10)
  --log_file_prefix=PATH           Path prefix for log files. Note that if you
                                   are running multiple tornado processes,
                                   log_file_prefix must be different for each
                                   of them (e.g. include the port number)
  --log_to_stderr                  Send log output to stderr (colorized if
                                   possible). By default use stderr if
                                   --log_file_prefix is not set and no other
                                   logging is configured.
  --logging=debug|info|warning|error|none 
                                   Set the Python log level. If 'none', tornado
                                   won't touch the logging configuration.
                                   (default info)

cache options:

  --cache_expiry_max               maximum cache expiry (seconds) (default
                                   7200)
  --cache_expiry_min               minimum cache expiry (seconds) (default 600)
  --mc_threads                     number of threads to use for Async Memcache
                                   (default 1)

format options:

  --default_format                 format to use for legacy / URLs (default
                                   html)

nbviewer options:

  --debug                          run in debug mode (default False)
  --frontpage                      path to json file containing frontpage
                                   content (default nbviewer/frontpage.json)
  --localfiles                     Allow to serve local files under
                                   /localfile/* this can be a security risk
  --no_cache                       Do not cache results (default False)
  --port                           run on the given port (default 5000)

provider options:

  --with_gitlab                    Enable the gitlab provider (default False)
  --without_dropbox                Disable the dropbox provider (default False)
  --without_gist                   Disable the gist provider (default False)
  --without_github                 Disable the github provider (default False)
  --without_url                    Disable the url provider (default False)

proxy options:

  --proxy_host                     The proxy URL.
  --proxy_port                     The proxy port.

rendering options:

  --processes                      use processes instead of threads for
                                   rendering (default 0)
  --threads                        number of threads to use for rendering
                                   (default 1)

ssl options:

  --sslcert                        path to ssl .crt file
  --sslkey                         path to ssl .key file

```

rgbkrk:
Would it be possible to make the prefix on the provider flags uniform?

```
provider options:

  --with_gitlab                    Enable the gitlab provider (default False)
  --with_dropbox                Enable the dropbox provider (default True)
  --with_gist                   Enable the gist provider (default True)
  --with_github                 Enable the github provider (default True)
  --with_url                    Enable the url provider (default True)
```

Then to call it,

```
python -m nbviewer --with-gitlab --with-github=False
```

bollwyvl:
Sure, that's simpler than the current setup. I just liked the simple
switches.

bollwyvl:
While I'm in there, i should probably make localfile a little more normal...

minrk:
I somehow didn't realize that we never got around to finish this one. Do you want to rebase and we can land it? I think we agreed upon most of the setup.

bollwyvl:
Oh, cool. I thought there were more outstanding issues. I'll take a look at
the rebase, shouldn't be that much!

bollwyvl:
As mentioned on the dev meeting, this needs pretty much a full re-write: this is not sad :) should be lots of lessons learned!

Carreau:
@bollwyvl I'm going to tag that as `closed-pr`, close it to cleanup the queue, but anyone feel free to reopen it if you have plan/time to rework on it. 

Thanks a lot for starting that though, and it was a pleasure to have you on last week dev meeting !

bollwyvl:
Yeah: I still like the entrypoints, but if we're not using them anywhere
else, closing this is the right call. Thanks!

Issue #423, opened 2015-03-12
Test using tornado.testing


rgbkrk:
In our current testing setup, we're doing functional tests with no way to do mocks against GitHub itself. This ends up causing a lot of bad builds when things are actually fine. Since things can time out between requests and nbviewer, I this there's a mismatch.

Using [`tornado.testing`](http://tornado.readthedocs.org/en/latest/testing.html) should help do tests properly for nbviewer.

bollwyvl:
I like the look of [mock-server](https://github.com/tomashanacek/mock-server#format) for mocking a REST endpoint (or whatever, I guess).

We then just generate a serialized dump of our current tests, (maybe make that repeatable) and store it right along with the tests.

bollwyvl:
had a look through mock-server... didn't look very maintained. [github3.py](https://github.com/sigmavirus24/github3.py/tree/develop/tests), as was mentioned, rolled (several) of their own. This seems like something that could stand alone eventually, but either bringing github3.py in and using it directly, or cargo-culting could work.

rgbkrk:
The major issue is that we can't bring in github3.py in directly since we need async futures instead of blocking network calls. (if you're talking client)

If you're talking bringing in the mocks from github3.py to test against, that's another thing.

PR #455, opened 2015-05-13
Twiddle github provider to fully work with GHE.

mrterry:
When using the github provider with GHE, only the raw view works.  Most of the URLs are hard coded to be for `github.com` rather than `my.internal.gh.enterprisey.com`.  This PR (which should not be merged) show an example of how to twiddle the URLs to make it work.

This PR clearly shouldn't be merged.  It needs a way to inject the internal domain.  Also, tests, docs, etc.

Is first-class support for GHE enterprise something nbviewer wants?  If so what is the best way to get this in?  separate provider (named perhaps `ghe`) that inherits from `github`?  Get the base URL and friends from `GithubClientMixin`?  Other?

### DIFF
--- a/nbviewer/providers/github/handlers.py
+++ b/nbviewer/providers/github/handlers.py
@@ -40,7 +40,7 @@ PROVIDER_CTX = {
 parts = ['github', 'corp', 'enterprisey', 'com']
 parts = ['github', 'com']
 
-base_url = u'https://' + '.'.join(parts)
+BASE_URL = u'https://' + '.'.join(parts)
 domain_re = '\.'.join(parts[:-1])
 sfx_re = '\.' + parts[-1]
 domain_sfx = '\.'.join(parts)
@@ -99,7 +99,7 @@ class GitHubUserHandler(GithubClientMixin, BaseHandler):
                 url=repo['name'],
                 name=repo['name'],
             ))
-        provider_url = base_url + u"/{user}".format(user=user)
+        provider_url = BASE_URL + u"/{user}".format(user=user)
         html = self.render_template("userview.html",
             entries=entries, provider_url=provider_url, 
             next_url=next_url, prev_url=prev_url,
@@ -151,7 +151,7 @@ class GitHubTreeHandler(GithubClientMixin, BaseHandler):
         base_url = u"/github/{user}/{repo}/tree/{ref}".format(
             user=user, repo=repo, ref=ref,
         )
-        provider_url = base_url + u"/{user}/{repo}/tree/{ref}/{path}".format(
+        provider_url = BASE_URL + u"/{user}/{repo}/tree/{ref}/{path}".format(
             user=user, repo=repo, ref=ref, path=path,
         )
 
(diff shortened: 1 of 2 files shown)


rgbkrk:
I (personally) am not going to work on GitHub enterprise support, but I'm happy to bring it in. @bollwyvl has been working at extracting providers and if GHE really has to be separate, so be it. Later versions of GHE are said to support rendering Jupyter notebooks in the same way they're supported on GitHub right now, so I'd weigh that as much as you can. There will always be notebooks nbviewer can render that GitHub can't though (for security).

@bmabey - thoughts?

bollwyvl:
Yeah, for private installs we want to support ghe. And stash. But
nbviewer.j.o won't ever, so it's a question of scope. If it's really easy
(once i read your pr), we should just do it, though.

Whether this means we move the ghe machinery over to a new repo or keep
everything bundled is a good question... Do we want to be on the hook for
maintaining all these proprietary apis? We will always support public
github, so maybe its meant to be. For stash? Probably not. Gitlab? Maybe.

When entry_points hits, it will be easier to ask that question... I'm not
happy with providers fully yet.

mrterry:
I'm currently pitching nbviewer as the way we ought to share analysis inside the company and am happy to do the work.  I just don't want to overly uglify the code base.

As for waiting on GHE to do the work for me, I'm not holding my breath.  geojson and csv rendering were both released mid 2013 and are not in GHE.  Time will tell if notebook rendering is different.

mrterry:
Regarding support for proprietary APIs.  I'd expect github and GHE to be nearly the same.  The only differences I've seen are that GHE has no githubusercontent.com, and GHE adds query strings with a token parameter for raw views.  My changes are basically `s/github.com/base_url`

bollwyvl:
Great, changes look clean. I'll do a PR inspired by yours.

My issue with the proprietary APIs: not sure how we'll test them right now. @rgbkrk has been talking about Testing 2.0 in #423 (and elsewhere), where we will start spoofing github, and this would be a great place to start that.

@mrterry If you would help provide some testing support in the meantime against your instance, that would help tremendously. I'll update once I PR!

bollwyvl:
Oh: one reason I could imagine making GHE a separate provider (still bundled, for the time being) would be the ability to turn off public github and other providers and turn on only ghe... so:
```bash
nbviewer --without-github --without--gist  --without-url \
  --with-ghe GHE_URL=my.internal.gh.enterprisey.com GHE_TOKEN=<key>
```
I can't imagine wanting to escalate the number of in-house servers that were running something that was basically a fiewall hole. For that matter, I really wouldn't run nbviewer internally "inside" a single-sign-on experience!

mrterry:
@bollwyvl Happy to test on our instance.

running both github and ghe leads to some ambiguities in what the search bar does.  if i type in `mrterry` does that go to github.com/mrterry or ghe.com/mrterry?  My current expectation is that you get either github or ghe, but not both.  links directly to ipynb files should always help.

mrterry:
this is being done properly in #456

PR #456, opened 2015-05-14
[wip] Adding Github Enterprise

bollwyvl:
Based on #455, this adds Github Enterprise (`ghe`) as a provider... it's use case is different enough from public github, and it serves as a pattern for other public:tm:/hosted:copyleft: solutions:
- bitbucket/stash
- sourceforge/allura
- gitlab/gitlab

Here is an example of what it would look like to run `ghe`, and no other providers:

```bash
GITHUB_API_URL=https://api.my.internal.gh.enterprisey.com/ \
    GITHUB_HTML_URL=https://my.internal.gh.enterprisey.com/ \
    GITHUB_API_TOKEN=<whatever> \
    python -m nbviewer \
        --providers=nbviewer.providers.ghe \
        --provider-rewrites=nbviewer.providers.ghe
```

At present, it won't really be possible to use `ghe` and `github`, as they share the same keyspace... please chime in if this is an important feature It will just necessitate some more hacks.... however, if running an internal nbviewer, I would likely not have `github` (and maybe even `url`) enabled.

Adding appropriate tests will be challenging, and should likely begin the investigation of #423, and mocking up "real" github (refreshably scraped) for faster tests.

@mrterry I'm still working on this! had to knock off for a while!

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ You must also specify your `OAUTH` or `API_TOKEN` as explained above.  For examp
 ```shell
 $ docker run -p 8080:8080 -e 'GITHUB_OAUTH_KEY=YOURKEY' \
                           -e 'GITHUB_OAUTH_SECRET=YOURSECRET' \
-                          -e 'GITHUB_API_URL=https://ghe.example.com/api/v3/' \
+                          -e 'GITHUB_API_URL=https://ghe.example.com/api/v3' \
                           jupyter/nbviewer
 ```
 
@@ -70,7 +70,7 @@ $ docker run -p 8080:8080 nbviewer
 
 The Notebook Viewer uses `memcached` in production. To locally try out this
 setup, a [docker-compose](https://docs.docker.com/compose/) configuration is
-provided to easily start/stop the `nbviewer` and `memcached` containers 
+provided to easily start/stop the `nbviewer` and `memcached` containers
 together from a your current branch. You will need to install `docker` prior
 to this.
 
@@ -124,7 +124,7 @@ This will automatically relaunch the server if a change is detected on a python 
 #### Running the Tests
 
 `nose` is used to run the test suite. The tests currently make calls to
-external APIs such as GitHub, so it is best to use your Github API Token when 
+external APIs such as GitHub, so it is best to use your Github API Token when
 running:
 
 ```shell
@@ -180,6 +180,6 @@ Formats are ways to present notebooks to the user.
 - `slides`
 
 #### Writing a new Format
-If you'd like to write a new format, open a ticket, or speak up on [gitter][]! 
-We have some work yet to do to support your next big thing in notebook 
+If you'd like to write a new format, open a ticket, or speak up on [gitter][]!
+We have some work yet to do to support your next big thing in notebook
 publishing, and we'd love to here from you.
--- a/nbviewer/app.py
+++ b/nbviewer/app.py
@@ -186,7 +186,7 @@ def main():
             max_cache_uris.add('/' + link['target'])
 
     fetch_kwargs = dict(connect_timeout=10,)
-    if options.proxy_host: 
+    if options.proxy_host:
         fetch_kwargs.update(dict(proxy_host=options.proxy_host,
                                  proxy_port=options.proxy_port))
 
--- a/nbviewer/providers/ghe/__init__.py
+++ b/nbviewer/providers/ghe/__init__.py
@@ -1,0 +1,4 @@
+from .handlers import (
+    default_handlers,
+    uri_rewrites,
+)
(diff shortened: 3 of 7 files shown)


mrterry:
nifty.  I'll point it at our GHE and see how things shake out.

mrterry:
works:
* /ghe/mrterry (list of repos)
* /ghe/mrterry/notebooks (view repo)

broken:
* /from /ghe/mrterry click repo name (404 @ /github/mrterry/notebooks/tree/master/)
* url to file (/mrterry/notebooks/blob/master/kdtree_grid.ipynb)

odd:
* GITHUB_HTTP_URL must have a trailing slash
* raw url goes to /urls/url (/mrterry/notebooks/master/kdtree_grid.ipynb?token=AAAA)

bollwyvl:
Hey, thanks for the tests!

_Something_ worked, which means we're at least not on the totally wrong track.

Thanks for the catches... it's feels a little weird, but you could PR against my PR... actually ends up as a very nice history/narrative. Anyhow, I've changed the basic stuff

> GITHUB_HTTP_URL must have a trailing slash

Yeah, that's the way I found API_URL... i don't feel strongly either way. So I would feel fine about normalizing that now, if that's more... normal.

bollwyvl:
Now `GITHUB_HTTP_URL` and `GITHUB_API_URL` do not require trailing slash. 

As for:
> broken
    - /from /ghe/mrterry click repo name (404 @ /github/mrterry/notebooks/tree/master/)
    - url to file (/mrterry/notebooks/blob/master/kdtree_grid.ipynb)

I can't reproduce any of the issues running with:
```
GITHUB_API_TOKEN=XXX python -m nbviewer \
  --debug \
  --no-cache \
  --providers=nbviewer.providers.ghe
```
but it's not a good test. Can you give us more detail? Logs, etc. Perhaps redacted, pasted into a gist?

mrterry:
i was going to do the PR against a PR, but this was easier.

On a separate note, as a consequence of the refactoring on the github provider,

mrterry:
I should have clarified, with the inheritance order fixed, everything seems to work.

The only bit of weirdness I've found is what happens when you have both the `github` and `ghe` providers in `default_providers`.  To isolate the behavior of `ghe`, i've been replacing `github` with `ghe`.

mrterry:
More detail on `ghe` and `github` competition.

For `ghe` to work, you pass in environment variables for auth tokens and such.  These get used by both the `github` and `ghe` clients.  This leads to the confusing situation where you pass in a username and `github` happily resolves from the `ghe` site.

I probably won't get to this before monday, but things might go smoother if the `github` and `ghe` providers looked for separate environment variables for auth and such.

bollwyvl:
great! I'll figure out some env scheme to clear up the competition!

mrterry:
see my PR (https://github.com/bollwyvl/nbviewer/pull/1) for a env scheme and to fix the broken tests.

bollwyvl:
@mrterry Great! I'll check this out! Sorry for the delay: my availability is not great right now.

We do need to get on the testing bit for ghe, no tests, no :ship: (though you'd be free to use the fork if need be). I'll have a look into the things mentioned in #423. Once we have some kind of plan for the API mocking, I may need your help with some sample data (probably some JSON/curl dumps)... redacted, obviously.

mrterry:
What level of testing is needed for a merge?

Mocking out the network requests would be good. Testing the redirects would
be good. Testing that URLs make the right network requests would be good.
I have some time to work on this.  Got any guidance on priority?

bollwyvl:
Yeah, mostly url testing. Our current stuff is pretty black box, based on
turning a URL into html. This isn't wrong, but we should be doing more.
Errors, etc.

We'll also want to test the interplay between ghe and GitHub, as you have
been working on.

I guess coverage, which we haven't been tracking, is the correct measure of
how much test... Mixins are trouble for coverage, though :(

I'll think some about it, but am open to proposals! I know @rgbkrk has
recused himself from this feature, but testing goal insights would be
relevant for this and several other issues.

bollwyvl:
condensed some of the environment variable noise with a prefix. just noticed moving the html url out of there... I agree that it isn't used for requests, but the data `html_url` that comes back will be in terms of that URL. Also, it would remove even more VARIABLE_NAMES from the scope.

rgbkrk:
Still WIP?

mrterry:
I'm running some variant on this internally. I never got around to writing
all the tests. Actually, making this accessible for tests would require
some nontrivial refactoring in nbviewer.

Carreau:
@bollwyvl same as other PR, tagging as closed-pr which is a bit stale, closing to clear-up the queue, but feel free to reopen as soon as you have time to rework on it. 

Thanks !

PR #459, opened 2015-05-19
enable mathjax equation numbering by default

greole:
It would be great if the nbviewer could handle numbered mathjax equations.

Check http://stackoverflow.com/questions/18823779/ipython-notebook-and-mathjax-labeled-equations
for more details

### DIFF
--- a/nbviewer/app.py
+++ b/nbviewer/app.py
@@ -87,6 +87,7 @@ def main():
     define("proxy_port", default="", help="The proxy port.", type=int)
     define("providers", default=default_providers, help="Full dotted package(s) that provide `default_handlers`", type=str, multiple=True, group="provider")
     define("provider_rewrites", default=default_rewrites, help="Full dotted package(s) that provide `uri_rewrites`", type=str, multiple=True, group="provider")
+    define("mathjax_url", default="https://cdn.mathjax.org/mathjax/latest/", help="URL base for mathjax package", type=str)
     tornado.options.parse_command_line()
 
     # NBConvert config
@@ -214,6 +215,7 @@ def main():
         render_timeout=20,
         localfile_path=os.path.abspath(options.localfiles),
         fetch_kwargs=fetch_kwargs,
+        mathjax_url=options.mathjax_url,
     )
 
     # handle handlers
--- a/nbviewer/formats.py
+++ b/nbviewer/formats.py
@@ -9,6 +9,7 @@ import re
 
 from IPython.nbconvert.exporters.export import exporter_map
 
+
 def default_formats():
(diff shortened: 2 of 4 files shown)


bollwyvl:
I agree in principal. Here on nbviewer, We should move mathjax config to it's own block so that slides, etc. can use it.

Upstream, as has been discussed, configuring mathjax at the metadata level would let people add numbering, or disable math altogether...

minrk:
Equation numbering is problematic in the live notebook because the number increments based on render order, rather than order on the page, so you get weird numbers. Nbviewer shouldn't have this problem, so I think it makes sense to enable it. The only peculiarity, then, is that it's a mismatch with the live notebook.

bollwyvl:
Great. I'll jigger the blocks so this happens in other formats. Might be a
nice opportunity to do a mathjax url config as well.

rgbkrk:
Fair points.

bollwyvl:
Jiggering on greole/nbviewer#1.

damianavila:
I am little worried about the fact that people will see numbering in nbviewer and then they will push for numbering working OK in the live notebook...

greole:
That is a valid point. However writing scientific reports without numbering is nearly impossible. I could live with a live notebook where I have to turn on numbering myself, but sharing a notebook with colleagues via nbviewer is a big issue. 

Currently to get numbering on nbviewer I use

    from IPython.display import HTML
    js = r""" <script type="text/Javascript"> MathJax.Hub.Config({ TeX: { equationNumbers: { autoNumber: "AMS", useLabelIds: true } }});
    </script> """
    HTML(js)
 
in a code cell. Couldn't this be simplified to something like

    from IPython.external.mathjax import enable_nbviewer_numbering 

which just spits out the corresponding js. That way it is possible to enable numbering in a simple, non disturbing way and turned off until explicitly demanded.

cstarknyc:
The js hack (Greole's) works nicely for an IPython notebook in Jupyter, but won't and doesn't work when rendered in GitHub. The practical effect is that communication with colleagues is compromised, since only those with Python set up can see the math line numbers.

bollwyvl:
> only those with Python

Or a web browser and, you know, nbviewer.org... Or perhaps i misunderstand
your comment.

cstarknyc:
Private GitHub repo, so as far as I know nbviewer can't do it's thing.

bollwyvl:
Indeed!

This use case is supported by hosting your own nbviewer and having a Github
api key for a robot which has access to those repos. This allows you any
kind of customization that your team may need, without making your host
your own github.

PR #464, opened 2015-06-08
Minor reorg for easier testing

mrterry:
Minor reorg to make unit testing providers easier.  Will get to #423 in time.

Makes uri translation function more accessible for tests and adds first pass for a github client that mocks out network communication.

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -1,6 +1,6 @@
 coverage
 invoke
-mock
+mock>=1.3.0
 nose
 requests
 -e .
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -1,6 +1,6 @@
 coverage
 invoke
-mock>=1.3.0
+mock>=1.3.0  # python34 and older versions of mock do not play well together.
 nose
 requests
 -e .
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -1,5 +1,6 @@
 coverage
 invoke
+mock
 nose
 requests
 -e .
--- a/nbviewer/providers/github/handlers.py
+++ b/nbviewer/providers/github/handlers.py
@@ -308,25 +308,22 @@ def default_handlers(handlers=[]):
 
 def uri_rewrites(rewrites=[]):
     return rewrites + [
-        (r'^https?://github\.com/([^\/]+)/([^\/]+)/tree/([^\/]+)/(.*)',
-            '/github/{0}/{1}/tree/{3}/{4}'),
-        (r'^https?://github\.com/([^\/]+)/([^\/]+)/tree/([^\/]+)/(.*)',
-            '/github/{0}/{1}/blob/{3}/{4}'),
-
         # three different uris for a raw view
         (r'^https?://github\.com/([^\/]+)/([^\/]+)/raw/([^\/]+)/(.*)',
-            '/github/{0}/{1}/blob/{2}/{3}'),
+            u'/github/{0}/{1}/blob/{2}/{3}'),
         (r'^https?://raw\.github\.com/([^\/]+)/([^\/]+)/(.*)',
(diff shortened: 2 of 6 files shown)


mrterry:
@bollwyvl This is nearly ready to go.  The tests pass on py2.7 and py3.3.  Something in mock on py3.4 is unhappy and I can't reproduce locally.  It looks like github is throttling me now, so tests are failing.

mrterry:
It looks like bumping the version of mock fixed the previous failure.  Github started throttling me on the last test, but all the github client tests (which use mock) passed.

bollwyvl:
Sounds good! I'll give it a look this weekend!

Really need to get our story together on not relying on GH for most of the
tests... i hate submodules from the PR perspective, but I would hate more
to have to embed xxMB of notebooks to not lose our "tribal knowledge".
probably need to do an instrumentation run and generate a list of all the
things we need.

mrterry:
this PR is a decent start.  you can't end-to-end test everything, but you should be able to fully test each layer.  As I see it there are 4-ish layers:

1. github client
1. the url regex
1. the handler
1. the template rendering

This add decent coverage for the github client and a subset of the url regexes.  In a follow up, I'll add test for some of the handlers.  I'll leave template rendering to someone else.

dev-requirements.txt
```diff
@@ -1,5 +1,6 @@
 coverage
 invoke
+mock>=1.3.0
```
bollwyvl: Should we just pin everything right now?

nbviewer/providers/github/handlers.py
```diff
@@ -294,8 +294,14 @@ def default_handlers(handlers=[]):
     """Tornado handlers"""
 
     return [
+        # ideally these URIs should have been caught by an appropriate
+        # uri_rewrite rather than letting the url provider catch them and then
+        # fixing it here.
+        # There are probably links in the wild that depend on these, so keep
+        # these handlers for backwards compatibility.
```
bollwyvl: Good summary

nbviewer/utils.py
```diff
@@ -8,6 +8,7 @@
 import base64
 import cgi
 import re
+from collections import OrderedDict
```
bollwyvl: A brave coder. I have tried to stay out of this file!

dev-requirements.txt
mrterry: This may be cavalier, but I don't see the need for dev requirements.  The mock requirement is to avoid a bug in mock that affects python 3.4.  I should add a comment describing this.

nbviewer/utils.py
mrterry: There be dragons here.  I have another change in the pipeline that will get the global state out of this module.

bollwyvl:
+1: Having some issues getting it (or anything) running with Docker (just did `fedup` and a few things aren't 100%), so I haven't tested it out there...

mrterry:
cool.  let me know if i can help with anything.  i'm eager to get these changes in so that I can get back to work on the GHE provider.  I have a version deployed internally, but there are few bugs related to gists that will greatly benefit from this.

bollwyvl:
Glad to hear your internal hosting is going well. My group still trying to position ourselves into getting to build a Stash integration.

Perhaps @Carreau or @rgbkrk have a look at this?

mrterry:
Gists need some work and there are still a few regex corner cases, but is
going well.

Carreau:
Well, I never say no to plenty of test.
Also I doubt I still have a huge expertise on nbviewer. 

+1 as far as I can tell.

nbviewer/utils.py
rgbkrk: Hahaha, omg.

rgbkrk:
I say :shipit:

Issue #473, opened 2015-06-24
localfile directory traversal


bburky:
The localfile provider allows directory traversals. A root directory is accepted as a command line argument to nbviewer, but requests are not prevented from escaping this root.

For example, when running using the Docker configuration and the `-localfiles=/srv/nbviewer/nbviewer/static` argument added to the nbviewer command, it is possible to access the notebook at `/srv/nbviewer/tests/notebook.ipynb` using the URL http://docker.local:8080/localfile/%2e%2e/tests/notebook.ipynb

Use curl to test this (Chrome likes to remove `..` and `%2e%2e` from URLs):

```
$ curl http://192.168.99.100:8080/localfile/%2e%2e/tests/notebook.ipynb
```

The following is logged by nbviewer:

```
[I 150624 04:37:57 handlers:36] looking for file: '/srv/nbviewer/nbviewer/static/../tests/notebook.ipynb'
[I 150624 04:37:57 base:505] rendering 287 B notebook from ../tests/notebook.ipynb
[I 150624 04:37:57 log:53] 200 GET /localfile/%2e%2e/tests/notebook.ipynb (192.168.99.1) 157.38ms
```

I realize localfile is documented as "this can be a security risk", but it isn't sandboxed to the specified path. It could probably use a stronger warning too.

Carreau:
Yeah, localfile is ment roughly only for debug purposes, like when you are fixing bug in a train or an a plane. Should be more explicit about that.

do you want to make a PR that use log.critical if people use localfile ?

pelson:
> I realize localfile is documented as "this can be a security risk", but it isn't sandboxed to the specified path. It could probably use a stronger warning too.

Incidentally, I stood up an internal nbviewer over a year ago before localfile existed, and one of the goals of that was to view all files outside of the specified path. Just wanted to point out that that is a desirable thing sometimes, but I agree, a more specific warning might be a good idea here.

Carreau:
I see the need for localfile, and it is oft requested, Though, if localfile
you have, you can likely nbconvert locally and have access to last modif
date-time to be much more efficient. so --localfile is probably no well
suited.

bollwyvl:
While working on provider entry points (need to revisit, might be ready for
review), i added path checking to chain it to a location... URLs with ".."
in them are never, ever good in my mind.

bollwyvl:
Fix deployed (not that production uses localfile anyway....)

Issue #472, opened 2015-06-24
XSS in GitHub Gist file not found error message


bburky:
The error message for files not found in a gist includes the missing filename to be included in the 404 response. This filename is unsanitized for HTML and enables an XSS. 

Chrome's XSS auditor will hide this simple XSS, look in the devtools console to see the error.

http://nbviewer.ipython.org/gist/bburky/7763555/%3Cimg%20src%3Dx%20onerror%3Dalert%281%29%3E

See also #471

Carreau:
For information this has been posted after the OP did report the security issue on security@ipython.org, which is the right thing to do. As the point of nbviewer is to allow arbitrary html/js this was not considered as needing to stay private. Hence this issue.

bollwyvl:
Thanks for the issue! There are likely other places where things are not
escaped correctly... I suppose we can't hide behind, "but there point is
_enabling_ xss!" On everything.

bburky:
Yeah, there are some other escaping issues. This same bug I think has weird behavior with `%0a`  (a newline) getting uneacaped and producing multiple lines of output in the log.

And a separate bug I forgot to report: go to the directory listing linked in #471 and click the "filetypes" link to get a 404 error. That branch name has a slash in it which gets accidentally unescaped.

bollwyvl:
The slash situation is another story, #411 :) But would certainly like to
knock out, "people might make bad urls on <provider>" in a consistent way

rgbkrk:
As I said in a private thread, I'd suggest starting with setting `autoescape=True` in the jinja2 environment:

https://github.com/jupyter/nbviewer/blob/7c530ee028c78ce522be4c1f422f47775d364460/nbviewer/app.py#L158

bollwyvl:
Fix deployed!

Issue #471, opened 2015-06-24
XSS in GitHub directory listings


bburky:
nbviewer allows browsing directory listings of the files in GitHub repositories. These directory listings do not properly sanitize the filenames and enable an XSS attack when rendered into HTML.

The the following URL and GitHub repository demonstrate the issue:
http://nbviewer.ipython.org/github/bburky/xss/tree/%3Cscript%3Ealert(1)%3C%2fscript%3E/
https://github.com/bburky/xss

See also #472

Carreau:
Thanks for the report.

For information this has been posted after the OP did report the security issue on security@ipython.org, which is the right thing to do. As the point of nbviewer is to allow arbitrary html/js this was not considered as needing to stay private. Hence this issue.

bburky:
The same issue also exists in the directory listing of gists too:

http://nbviewer.ipython.org/gist/bburky/c020825874798a6544a7

bollwyvl:
Fix deployed!

PR #474, opened 2015-06-26
Template security

bollwyvl:
This adds tests and fixes for #471, #472 and #473.

As per @Carreau's suggestion, now using `autoescape` on the jinja2.Environment.

The localfile handler also will only serve files found under its path.

if someone really wants to fight for being able to allow escaping the directory they specify, i propose we put it behind a big gnarly config flag, i.e. `--very-insecure-traversal`.

I looked into a few tools for broader XSS exploration, but i think we'll have a lot of false positives... also, don't know which of them i can trust.

### DIFF
--- a/nbviewer/tests/test_security.py
+++ b/nbviewer/tests/test_security.py
@@ -1,0 +1,37 @@
+# coding: utf-8
+#-----------------------------------------------------------------------------
+#  Copyright (C) 2015 The Jupyter Development Team
+#
+#  Distributed under the terms of the BSD License.  The full license is in
+#  the file COPYING, distributed as part of this software.
+#-----------------------------------------------------------------------------
+
+import requests
+
+from .base import NBViewerTestCase
+
+from ..providers.local.tests.test_localfile import (
+    LocalFileDefaultTestCase as LFDTC
+)
+
+class XSSTestCase(NBViewerTestCase):
+    def _xss(self, path, pattern='<script>alert'):
+        r = requests.get(self.url() + path)
+        self.assertEqual(r.status_code, 200)
+        self.assertNotIn(pattern, r.content)
+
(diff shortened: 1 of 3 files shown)


Carreau:
Was it my suggestion to use autoescape ? I don't remmeber.

Carreau:
+1

Carreau:
btw, are you on the security@ipython ML?

bollwyvl:
My bad, @rgbkrk suggested autoescape! Luckily we weren't building any ui with string manipulation, so it was pretty much a win.

And I'm not on the security council...

rgbkrk:
Awesome, thank you!

nbviewer/tests/test_security.py
```diff
@@ -0,0 +1,37 @@
...
+        self.assertEqual(r.status_code, 200)
+        self.assertNotIn(pattern, r.content)
+
+    def test_github_dirnames(self):
+        self._xss(
+            '/github/bburky/xss/tree/%3Cscript%3Ealert(1)%3C%2fscript%3E/'
+        )
+
+    def test_gist_filenames(self):
+        self._xss('/gist/bburky/c020825874798a6544a7')
```
rgbkrk: Ooh tests! Yay!

bollwyvl: Yeah... Seems somewhat lacking, but how much further do we go? Red team docker container?

bollwyvl:
It might be an "exciting" deployment... I'm offline for the next 48 so maybe wait until Sunday?

rgbkrk:
Well, it'll automatically happen on its own when CoreOS goes through an upgrade cycle...

bollwyvl:
Hooray automation! Need to learn some core....

PR #477, opened 2015-07-09
Included two additional dependencies

agartland:
I was following the install directions on a fresh install of Ubuntu and Anaconda and these additions were neccessary.

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -2,4 +2,6 @@ coverage
 invoke
 nose
 requests
+futures
+markdown
 -e .


agartland:
Apologies, I see that these depencies are in the requirements.txt file (which I should have already run).
I did need to install the `node` package and run `npm install` to get all the steps to work though. Now it seems to be working (in <30 min thanks to great documentation!). The CSS doesn't quite look right, but I'll keep poking around. Thanks for this great project!

bollwyvl:
have a look at the dockerfile or travis.yml: we have wrapped up the node
stuff in invoke, which is part of dev-requirements.txt

once we do a Real Release, we'll ship the static assets, so pip installing
won't require node at all.

agartland:
Thanks. Heading over to gitter to ask about the CSS...

Issue #480, opened 2015-07-22
Dropbox links for notebooks stopped working


cchwala:
The problem seems to be that nbviewer.ipython.org does something wrong when trying to get the "raw download" URL from the dropbox link I usually provide.

Example:
Working nbviewer link from maybe two month ago
http://nbviewer.ipython.org/urls/dl.dropbox.com/s/p9qyjmv06wglqsx/Read%20CSV%20data%20for%20Abidjan.ipynb?dl=0

Dropbox link from some minutes ago
https://www.dropbox.com/s/p9qyjmv06wglqsx/Read%20CSV%20data%20for%20Abidjan.ipynb?dl=0

Letting nbviewer do its job, I get
http://nbviewer.ipython.org/urls/dl.dropbox.com/s/p9qyjmv06wglqsx/Read%2520CSV%2520data%2520for%2520Abidjan.ipynb/%253Fdl%253D0
and a HTTP 403: Forbidden

The differences seem just to be the URL path encondings of the special characters (?, =, space).

BjornFJohansson:
I have a problem that I think is caused by the same issue. I also serve IPyhton notebooks
from dropbox (this is a great way to share notebooks). I have local files in the same folder
as the notebook and I link to them in markdown cells:

    [file](file.ext)

This worked until a couple of days ago. See [example](http://nbviewer.ipython.org/urls/dl.dropboxusercontent.com/u/1263722/local_link2.ipynb?flush_cache=true)

This breaks nbviewer for me! Is this a bug or a feature?

bollwyvl:
Curious indeed. I'll have a look into it, though probably this evening.

BjornFJohansson:
Thanks! I am about to submit a paper that depends an app that produces a folder of notebooks
with links to files contained in the folder and other notebooks.

satra:
same issue here.

rgbkrk:
Hmmm, that's looking like double encoding to me.

bollwyvl:
Yes, there is fairly naive url_encoding occuring in base that cam in from #421. Having Travis take a look, seeing what's going on...

bollwyvl:
Yep, that was in there to fix unicode encoding issues which crop up in python 3. At this point, I think it's just in the tests, as the browsers seem to do the Right Thing... but not making tremendous headway.

BjornFJohansson:
I use chrome on Linux, Firefox on windows show the same problem.
Thanks for the time you put in to this!

[...]

Issue #379, opened 2014-12-11
Encoded colon causes issues in rendering notebooks on arbitrary port


rgbkrk:
As reported in #297, we are currently encoding the colon in the port portion of the URL.

bollwyvl:
Should be fixed now!

PR #481, opened 2015-07-25
Fixing URL double-encode, dropbox-specifics

bollwyvl:
This adds some more robust redirect URL handling with `urlunparse`.

As it was the source of the problem reported in #480, specifically for dropbox, it also strips the `?dl=.` which the upstream doesn't care about anyway. To allow that, the `?` param handling now occurs _after_ `url_rewrites`.

While I was in there testing, I handled #379, I think, as it was messing up my local test stuff (since I don't have/want/trust dropbox).

Would love some other eyes on this, as we should get it out pretty soon!

### DIFF
--- a/nbviewer/providers/base.py
+++ b/nbviewer/providers/base.py
@@ -17,10 +17,10 @@ from datetime import datetime
 try:
     # py3
     from http.client import responses
-    from urllib.parse import urlparse
+    from urllib.parse import urlparse, urlunparse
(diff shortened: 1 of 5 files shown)


mcs07:
This looks good to me - it fixes some issues I've been having with the escaping of both port numbers and query strings.

rgbkrk:
Sweet.

Issue #427, opened 2015-03-18
nbviewer can't render files in Google Drive


Sisyphuss:
Nbviewer failed to render a valid Julia notebook:
  https://drive.google.com/file/d/0B30Cah_MN0i0TlZfRlBXLWgxbE0/view?usp=sharing
stored in Google Driver.

bollwyvl:
Well, that link doesn't resolve to a notebook, but rather a download page. That's not going to work.

Pulling the URL that actually gets generated from the download  (actually annoying to get, as it is all javascripty, but at least stable) yields:

https://doc-00-ag-docs.googleusercontent.com/docs/securesc/ha0ro937gcuc7l7deffksulhg5h7mbp1/l058tmm4stdqdjhe1hs0t564snj08ahm/c?e=download

And feeding that into the front page of nbviewer, we get the nbviewer URL:
http://nbviewer.ipython.org/urls/doc-00-ag-docs.googleusercontent.com/docs/securesc/ha0ro937gcuc7l7deffksulhg5h7mbp1/l058tmm4stdqdjhe1hs0t564snj08ahm/1426680000000/09083926091300078733/*/0B30Cah_MN0i0TlZfRlBXLWgxbE

we are getting a 403: forbidden from the upstream provider.

Checking this with curl:
```bash
curl "https://doc-00-ag-docs.googleusercontent.com/docs/securesc/ha0ro937gcuc7l7deffksulhg5h7mbp1/l058tmm4stdqdjhe1hs0t564snj08ahm/1426680000000/09083926091300078733/*/0B30Cah_MN0i0TlZfRlBXLWgxbE0?e=download"
```
Shows the URL will actually reply, but just not to nbviewer.ipython.org.

So basically, we would either need:
- google to allow nbviewer to request unauthenticated files
  - which they are unlikely to do, and would not be ultimately very useful, because a notebook without related assets is not very useful, and at present I doubt that any associated images, etc. would be addressable by a normal URL scheme.
- a new provider for google drive, and use the API itself to retrieve public files

The provider option seems like a good idea, along with #371 (gitlab), #302 (webgit), and #402 (stash/bitbucket). I've included this in #428!

dominikandreas:
Google Drive has worked for me in the past using the web host feature:
https://support.google.com/drive/answer/2881970?hl=en

You just need to extract the document id from the sharing link and use this address instead:
www.googledrive.com/host/[doc id] 

[...]

PR #482, opened 2015-07-28
adding google drive rewrite provider

bollwyvl:
A dropbox-style provider based on #427.

Can take the two kinds of links i found in drive (the "share" UI and the "get link" UI).

Tests passing currently blocked by #481, as it uses `?` params.

### DIFF
--- a/nbviewer/providers/__init__.py
+++ b/nbviewer/providers/__init__.py
@@ -9,7 +9,8 @@ default_providers = ['nbviewer.providers.{}'.format(prov)
                      for prov in ['url', 'github', 'gist']]
 
 default_rewrites = ['nbviewer.providers.{}'.format(prov)
-                    for prov in ['gist', 'github', 'dropbox', 'url']]
+                    for prov in ['gist', 'github', 'dropbox', 'googledrive',
+                                 'url']]
 
 
 def provider_handlers(providers=None):
--- a/nbviewer/providers/googledrive/__init__.py
+++ b/nbviewer/providers/googledrive/__init__.py
@@ -1,0 +1,1 @@
+from .handlers import uri_rewrites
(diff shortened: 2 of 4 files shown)


bollwyvl:
Turns out there is a way to publicly [host a whole directory](https://developers.google.com/drive/web/publish-site), which enables relative links:

1. This publicly-shared folder:
  - https://drive.google.com/folderview?id=0B7d519FxJIqPfkg0Wk1XSW42cWRQMnFTdXVLVkdUNWV0c0JRX00tVGpEblF3U1FqYnJjckE&usp=sharing
2. Can be dereferenced to this link:
  - http://googledrive.com/host/0B7d519FxJIqPfkg0Wk1XSW42cWRQMnFTdXVLVkdUNWV0c0JRX00tVGpEblF3U1FqYnJjckE/
3. Which contains a link to `Untitled.ipynb`:
  - https://4faab7555cc0ed4cb029e9ed361eb8e38c4b85fc.googledrive.com/host/0B7d519FxJIqPfkg0Wk1XSW42cWRQMnFTdXVLVkdUNWV0c0JRX00tVGpEblF3U1FqYnJjckE/Untitled.ipynb
4. The first part of the domain can be stripped to yield this valid nbviewer link:
  - http://nbviewer.ipython.org/url/googledrive.com/host/0B7d519FxJIqPfkg0Wk1XSW42cWRQMnFTdXVLVkdUNWV0c0JRX00tVGpEblF3U1FqYnJjckE/Untitled.ipynb

On my local branch with #481 applied, this actually shows the kitty.

So while that is an awful link, it is actually pretty viable for near-term integration, though it would probably require some documentation/screecasting/bookmarkleting, as it requires URL hacking... or index parsing, which I don't think is worth it. However, given that http://github.com/jupyter/jupyter-drive is a thing, this could be a very powerful way indeed to work on notebooks for publication with a team.

bollwyvl:
Let's not do this!

http://googleappsdeveloper.blogspot.com/2015/08/deprecating-web-hosting-support-in.html

rgbkrk:
Of course...

PR #486, opened 2015-08-02
Reorg2

mrterry:
This builds on #464 and removes globals state from the provider registration process.

### DIFF
--- a/nbviewer/handlers.py
+++ b/nbviewer/handlers.py
@@ -4,8 +4,6 @@
 #  Distributed under the terms of the BSD License.  The full license is in
 #  the file COPYING, distributed as part of this software.
 #-----------------------------------------------------------------------------
-from collections import OrderedDict
-
 from tornado import web
 from tornado.log import app_log
 
@@ -49,7 +47,7 @@ class CreateHandler(BaseHandler):
 
     only redirects to the appropriate URL
     """
-    uri_rewrite_dict = None
+    uri_rewrite_list = None
 
     def post(self):
         value = self.get_argument('gistnorurl', '')
@@ -60,15 +58,12 @@ class CreateHandler(BaseHandler):
     def get_provider_rewrites(self):
         # storing this on a class attribute is a little icky, but is better
         # than the global this was refactored from.
-        if self.uri_rewrite_dict is None:
+        if self.uri_rewrite_list is None:
             # providers is a list of module import paths
             providers = self.settings['provider_rewrites']
 
-            uri_rewrite_dict = OrderedDict()
-            uri_rewrite_dict.update(provider_uri_rewrites(providers))
-
-            type(self).uri_rewrite_dict = uri_rewrite_dict
-        return self.uri_rewrite_dict
+            type(self).uri_rewrite_list = provider_uri_rewrites(providers)
+        return self.uri_rewrite_list
 
 
 #-----------------------------------------------------------------------------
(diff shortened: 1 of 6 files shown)


mrterry:
@bollwyvl more refactoring to make writing tests easier.

bollwyvl:
Looks good... but we may want to wait to merge/rebase this one until #464 is in... looks like a lot of duplicate changes.

bollwyvl:
Ah, I see now that #464 is a strict subset, at least for the time being!

mrterry:
Yeah. This branch presupposes the other.

bollwyvl:
@mrterry thanks! please rebase this at your leisure... but I'll start it checking out!

nbviewer/handlers.py
```diff
@@ -45,12 +47,25 @@ class CreateHandler(BaseHandler):
...
+    uri_rewrite_list = None
+
     def post(self):
         value = self.get_argument('gistnorurl', '')
-        redirect_url = transform_ipynb_uri(value, self.provider_rewrites)
+        redirect_url = transform_ipynb_uri(value, self.get_provider_rewrites())
         app_log.info("create %s => %s", value, redirect_url)
         self.redirect(redirect_url)
 
+    @classmethod
```
bollwyvl: good stuff. did something very similar over on ipymd, though I think it used a singleton instance.

nbviewer/utils.py
```diff
@@ -53,40 +53,26 @@ def url_path_join(*pieces):
...
         if matches:
-            value = rewrite.format(*matches.groups())
-            break
+            return rewrite.format(*matches.groups())
 
     # encode query parameters as last url part
-    if '?' in value:
-        value, query = value.split('?', 1)
-        value = '%s/%s' % (value, quote('?' + query))
+    if '?' in uri:
```
bollwyvl: Perhaps using urlparse here? considered doing this when i was working on the dropbox links. Not that this is wrong, and clearer.

nbviewer/handlers.py
```diff
@@ -45,12 +47,25 @@ class CreateHandler(BaseHandler):
...
+    def get_provider_rewrites(cls):
+        # storing this on a class attribute is a little icky, but is better
+        # than the global this was refactored from.
+        if cls.uri_rewrite_list is None:
+            # providers is a list of module import paths
+            providers = self.settings['provider_rewrites']
```
bollwyvl: __cls__.settings!

bollwyvl: Ah, a bit more involved than that, since `settings` actually would only be on `self`. hmm...

bollwyvl: Perhaps this?
```python
    def get_provider_rewrites(self):
        # storing this on a class attribute is a little icky, but is better
        # than the global this was refactored from.
        if self.uri_rewrite_list is None:
            # providers is a list of module import paths
            providers = self.settings['provider_rewrites']
            setattr(self.__class__, "uri_rewrite_list",
                    provider_uri_rewrites(providers))
        return self.uri_rewrite_list
```

mrterry: i think

    type(self).uri_rewrite_list = providers

is easier to read.

bollwyvl:
Looks really good! I'll press the green button tomorrow evening, if anyone else wants to take a look at it, but these are all positive changes in my eyes.

bollwyvl:
Thanks again @mrterry! We really appreciate the contributions! Let me know how you want to proceed on #456. I'll also took a look at some of my outstanding stuff: it would be good to get ready to be able to cleanly move to the 4.0 line without a lot of PR debt!

minrk:
Awesome, thanks both of you!

PR #487, opened 2015-08-03
Remove Mathjax render on fragments-related events.

damianavila:
This is related with https://github.com/jupyter/nbconvert/pull/81

### DIFF
--- a/nbviewer/templates/formats/slides.html
+++ b/nbviewer/templates/formats/slides.html
@@ -88,8 +88,7 @@
 
         // show/hide the nbviewer header on slide change
         Reveal.addEventListener('slidechanged', update);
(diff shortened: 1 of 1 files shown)


damianavila:
I don't think the failure is related... but... maybe I am missing something?

bollwyvl:
Hm, can't recall why that would have been in there... what pain is it causing you? Can you link to a notebook that demonstrates why this is a good thing?

damianavila:
Check this comment: https://github.com/jupyter/nbconvert/issues/79#issue-98542872

bollwyvl:
Ah, thanks for the context.

I do worry about them resizing themselves, as mathjax does some dirty tricks. But if it's causing someone issues, let's go ahead and see if it bothers _someone else_!

PR #488, opened 2015-08-16
Fixed imports of deprecated packages related to IPython.

iamaris:
Running the current nbviewer gives you several warnings about importing packages that have been deprecated. It seems to be related to the move of the IPython notebook into Jupyter.

These were the messages I got:
/home/aris/nbviewer/nb/local/lib/python2.7/site-packages/IPython/config.py:13: ShimWarning: The `IPython.config` package has been deprecated. You should import from traitlets.config instead.
  "You should import from traitlets.config instead.", ShimWarning)

/home/aris/nbviewer/nb/local/lib/python2.7/site-packages/IPython/html.py:14: ShimWarning: The `IPython.html` package has been deprecated. You should import from `notebook` instead. `IPython.html.widgets` has moved to `ipywidgets`.
  "`IPython.html.widgets` has moved to `ipywidgets`.", ShimWarning)

/home/aris/nbviewer/nb/local/lib/python2.7/site-packages/IPython/nbformat.py:13: ShimWarning: The `IPython.nbformat` package has been deprecated. You should import from nbformat instead.
  "You should import from nbformat instead.", ShimWarning)

/home/aris/nbviewer/nb/local/lib/python2.7/site-packages/IPython/nbconvert.py:13: ShimWarning: The `IPython.nbconvert` package has been deprecated. You should import from ipython_nbconvert instead.
  "You should import from ipython_nbconvert instead.", ShimWarning)

I fixed it by importing the updated package instead of the deprecated version. I also tested it after 
making the changes.

PR #493, opened 2015-08-29
Use jupyter

bollwyvl:
This builds on #488, adding the Docker changes needed for deployment.

It is now:
- based on `debian/jessie`
- still uses pip
  - Q: should we use conda?
- installs `notebook`, and not just `nbconvert` as we use the static assets from notebook, as well as some ipython util functions
  - i assume the assets will be on npm/bower eventually...
  - i am tempted to install nbconvert master, where most of our upstream fixes will come from, so that we can take advantage of one of the goals of TBS without pulling the bleeding edge of everything

This should touch off a round of ticket closings as we get unbackported nbconvert fixes!


Dockerfile
```diff
@@ -1,37 +1,58 @@
...
+    nodejs \
+    nodejs-legacy \
+    npm \
+    pandoc \
+    python3-dev \
+    python3-pip \
+    sqlite3 \
+    zlib1g-dev \
+  && apt-get clean \
+  && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
```
rgbkrk: Looking good.

Dockerfile
```diff
@@ -1,37 +1,58 @@
...
 
 # To change the number of threads use
 # docker run -d -e NBVIEWER_THREADS=4 -p 80:8080 nbviewer
 ENV NBVIEWER_THREADS 2
 EXPOSE 8080
 
-RUN pip install invoke && \
-    pip3 install invoke
+RUN pip3 install invoke
```
rgbkrk: Yay for one python, the true python: Python 3! :snake:

Dockerfile
```diff
@@ -1,37 +1,58 @@
...
+# front-end dependencies
 ADD ["./nbviewer/static/bower.json", "./nbviewer/static/.bowerrc", \
      "/srv/nbviewer/nbviewer/static/"]
 RUN invoke bower
 
+# build css
 ADD . /srv/nbviewer/
 RUN invoke less
 
+# root up until now!
```
rgbkrk: Yeah, guess we should fix that later.

nbviewer/app.py
```diff
@@ -23,7 +23,7 @@
 
 from jinja2 import Environment, FileSystemLoader
 
-from IPython.config import Config
+from traitlets.config import Config
```
rgbkrk: :boom:

nbviewer/static/bower.json
```diff
@@ -8,7 +8,7 @@
     "moment": "~2.8.4",
     "bootstrap": "components/bootstrap#~3.3",
     "font-awesome": "~4",
-    "pygments": "2.0.0",
-    "reveal.js": "~3.0.0"
+    "pygments": "~2.0.0",
+    "reveal.js": "~3.1.0"
```
rgbkrk: Cool.

requirements.txt
```diff
@@ -1,10 +1,7 @@
-tornado>=3.1.1
-
-pycurl
-pylibmc
+elasticsearch
```
rgbkrk: You don't have to rip it out of here now, but we could toss elasticsearch in a follow on PR (then figure out what we're doing after).

rgbkrk:
Now we wait for Docker Hub!

bollwyvl:
tick tock tick tock.

thanks for the quick, thorough review and merge!

nbviewer/static/bower.json
damianavila: yep... now we are on 3.1.0 in both sides (nbviewer and nbconvert) :+1:

PR #496, opened 2015-09-02
Swap order of dev and requirements and dev

rgbkrk:
This was probably a leftover from #493, now that the name has changed for the dev requirements file.

/cc @bollwyvl


bollwyvl:
Yup. My bad.

rgbkrk:
No prob, pinged in case there was more we missed.

Issue #507, opened 2015-09-24
markdown internationalization in nbviewer


sandiego206:
is it possible to  display correctly spanish letters such as [ñ, á, é, í, ó, ú] in a notebook I pass to nbviewer?

Carreau:
As far as I can tell it works fine already : http://nbviewer.ipython.org/gist/Carreau/1757899baeac304a28f3

sandiego206:
it happens when a cell is in markdown

Carreau:
The example I link above has both a code cell and a markdown cell, which both works. . Why not linking to a failing example ?

sandiego206:
here is [one](http://nbviewer.ipython.org/urls/bitbucket.org/sandiego206/asdasd/raw/master/Untitled.ipynb) it's supposed to say "El niño está ahí"

Carreau:
Ah! It's a bitbucket specific issue !
Same notebook on [github gist](https://gist.github.com/Carreau/cf5936646059b7474a66).

Will see what we can do.

willingc:
Thanks @sandiego206 for the link.

@Carreau I made a gist (https://gist.github.com/willingc/5de6a2d7aeebde35ccc8) for the example that @sandiego206 has on bitbucket. If I enter that gist URL into nbviewer, it renders correctly.

<img width="740" alt="screen shot 2015-09-24 at 7 20 55 pm" src="https://cloud.githubusercontent.com/assets/2680980/10091537/8a19141c-62f1-11e5-8689-079247385e4b.png">

I actually think this may be a bitbucket issue. Looking at the example @sandiego206 posted, the  URL from the `nbviewer` page is http://nbviewer.ipython.org/urls/bitbucket.org/sandiego206/asdasd/raw/master/Untitled.ipynb

Working backwards, I am guessing that the URL that was entered into `nbviewer` was https://bitbucket.org/sandiego206/asdasd/raw/master/Untitled.ipynb. Putting that link in my Chrome browser, pulls up the following screenshot: 
<img width="754" alt="screen shot 2015-09-24 at 7 33 06 pm" src="https://cloud.githubusercontent.com/assets/2680980/10091684/423006fe-62f3-11e5-811e-c3392456823b.png">

[...]

PR #508, opened 2015-10-01
force utf-8 for notebook json requests

bollwyvl:
JSON (and therefore all ipynb) should be served as utf-8.

This proposes that the url provider always use `utf-8` to decode what is coming back.

The default of [`ISO-8859-1`](https://github.com/jupyter/nbviewer/blob/4df8cfd919c87e03c029638b0b060555cd340c51/nbviewer/utils.py#L98) when `text/*` is received could stand, but maybe even it should be `utf-8`?

### DIFF
--- a/nbviewer/app.py
+++ b/nbviewer/app.py
@@ -51,12 +51,14 @@ app_log = log.app_log
 here = os.path.dirname(__file__)
 pjoin = os.path.join
 
+
 def nrhead():
     try:
         import newrelic.agent
     except ImportError:
         return ''
     return newrelic.agent.get_browser_timing_header()
+
 
 def nrfoot():
     try:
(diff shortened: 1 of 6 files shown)


bollwyvl:
In response to #507 :)

damianavila:
Tests are failing but from a quick view I am not sure if it is related...

bollwyvl:
> Tests are failing but from a quick view I am not sure if it is related...

yep, just needed to rerun them....

I didn't add a test, or demonstrate a failing one... should probably do that...

damianavila:
> I didn't add a test, or demonstrate a failing one... should probably do that...

Yep

bollwyvl:
Sorry for the delay!

I've added a test case looking for `ñ` in `/urls/bitbucket.org/sandiego206/asdasd/raw/master/Untitled.ipynb`.

I've added a partial implementation of a simple use of tornado.testing, a la #423, but it is still going out to the network. Definitely thinking of how to do the real mocking, though!

bollwyvl:
Had to do some more refactoring of when options get initialized, parsed.

bollwyvl:
Ok, tests passing! Underlying question remains: is Latin-1 really what we want to default to in the absence of any other guidance, or should that be `utf-8`?

minrk:
@bollwyvl my thinking: since we know of at least one case of a server providing utf8 data without specifying, and we *don't* know of any such cases serving latin1, let's switch it to utf8 for now, and keep an eye out for encoding-related bug reports.

bollwyvl:
Changed the default to UTF-8... Travis gardening done... I think this is ready to go!

minrk:
Thanks!

PR #511, opened 2015-10-06
slide templates

### LINKED ISSUE #506: Restore Scrolling Slides. Again.
It's #433, #439 all over again. My selective memory prevents me from noticing non-scrolling slides as a bug.

This somehow got broken again likely when we moved over to jupyter with #493.

See [this comment](https://github.com/jupyter/nbviewer/issues/466#issuecomment-142063405)

bollwyvl:
This fixes #506 and removes the awful body regex that breaks with bokeh, etc.

To do this, nbviewer now specifies its own template path for nbconvert, which can override or import the base ones. We have much more control over what we get from nbconvert this way, but are duplicating some effort: however, updating to `master` whenever nbconvert 4.next lands will be less of a big deal than it would have been otherwise for slides.

This is branched this off #508, but probably shouldn't have been. Whoops.

cc @damianavila

### DIFF
--- a/nbviewer/app.py
+++ b/nbviewer/app.py
@@ -51,12 +51,14 @@ app_log = log.app_log
 here = os.path.dirname(__file__)
 pjoin = os.path.join
 
+
 def nrhead():
     try:
         import newrelic.agent
     except ImportError:
         return ''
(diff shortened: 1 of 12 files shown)


bollwyvl:
Wrong branch!

PR #516, opened 2015-10-15
GitHub url sanitizing

bollwyvl:
The github auth tokens leak in URL params on certain pages with next/prev links.

This doesn't really impact any confidential data: some joker might use up our quota, as I assume that token doesn't do anything else, but that has to be about it. Would need to refresh the keys anyway!

This might be a thing that should be pushed up into the providers, but this seemed like the most expedient place to fix it.

cc @Carreau @rgbkrk

### DIFF
--- a/nbviewer/tests/test_security.py
+++ b/nbviewer/tests/test_security.py
@@ -35,3 +35,14 @@ class LocalDirectoryTraversalTestCase(LFRPTC):
         url = self.url('localfile/../README.md')
         r = requests.get(url)
         self.assertEqual(r.status_code, 404)
+
+
+class URLLeakTestCase(NBViewerTestCase):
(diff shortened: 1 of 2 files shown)


rgbkrk:
Ha, awesome. Thanks.

Carreau:
+1, need to have an issue to actually remember to track where this is coming from and prevent the param to be in the URI in the first place ?

bollwyvl:
We're reading the response headers, and finding the `Link`s, making them
into client links.

https://developer.github.com/v3/#oauth2-keysecret

They explicitly mention not letting it leak out... and the warning has been
there for some time.

minrk:
Thanks, we can reset the secret, key on the next deployment.

PR #530, opened 2015-11-04
Add the ability to specify custom templates

jdfreder:


nbviewer/app.py
```diff
@@ -136,10 +136,14 @@ def make_app():
 
     # setup tornado handlers and settings
 
-    template_path = pjoin(here, 'templates')
+    template_paths = pjoin(here, 'templates')
+    custom_template_path = os.environ.get('CUSTOM_TEMPLATE_PATH', None)
```
minrk: probably should use NBVIEWER in env variable names

minrk: ...and can probably leave out CUSTOM, e.g. NBVIEWER_TEMPLATE_PATH

jdfreder: :heavy_check_mark: - latest commit

Carreau:
I think we might want to rely on something more general like https://github.com/jupyter/nbconvert/pull/141 that allow to load template from other packages/location (that might not be disks). But i'm fine with that in the meantime.

jdfreder:
Not sure why the tests are failing here, does that mean anything?  Tested locally and it still works as expected even when the env var isn't defined. 



> I think we might want to rely on something more general like jupyter/nbconvert#141 that allow to load template from other packages/location (that might not be disks). But i'm fine with that in the meantime.

Yup

Carreau:
> Not sure why the tests are failing here, does that mean anything? Tested locally and it still works as expected even when the env var isn't defined.

https://github.com/jupyter/nbviewer/blob/0eeff0cd48acb894685b6edcd67796fde61adecb/nbviewer/formats.py#L61-L63 ?

Because template_path is a **nbconvert** api ?

jdfreder:
> Because template_path is a nbconvert api ?

No, unfortunately, that's unrelated to the code here.  Let me see if the tests are failing on master...

jdfreder:
Looks like the the tests in question are indeed failing in master: https://github.com/jupyter/nbviewer/pull/531

jdfreder:
@minrk did you remove your gist /gist/minrk/7518294/Untitled0.ipynb  ?

jdfreder:
@minrk interesting, I wonder if this is a problem on GitHub's end.  https://gist.github.com/minrk/7518294 is still there, but the notebooks aren't rendering.  I'm seeing "something went wrong".  Trying to access them directly yields a 404.

jdfreder:
... probably because the format is out of date.  Still, I'm not sure why this is causing nbviewer to hang, which should be pulling from RAW, right?

bollwyvl:
There are indeed two template environments:
- [app.py's](https://github.com/jupyter/nbviewer/pull/530/files#diff-a11c4b463a683e9ae44fa2ce29306d23L142) is the nbviewer app: home page, gist list, etc.
- [exporter's](https://github.com/jupyter/nbconvert/blob/00251fb42ca8665ba3d6faf8d531fee235bca6fd/nbconvert/exporters/templateexporter.py#L291) is the nbconvert app, and actually gets loaded for rendering notebooks

So: if you want to make pretty a branded nbviewer, you'd hack on the first one, and to get notebooks to be pretty, you'd use the second. What are you looking to do? Do we need both, as in `template_path` and `nbconvert_template_path`?

As to the tests: it's just "normal" API quota exhaustion... we really need to stub those out with fixture data!

jdfreder:
> What are you looking to do? Do we need both, as in template_path and nbconvert_template_path?

I just need the first one, but it would make sense to eventually have support for the second one.

bollwyvl:
My only criticism, would be that convention currently in app.py would be to expose `--template_path` as a [command line option](https://github.com/jupyter/nbviewer/blob/master/nbviewer/app.py#L225). I am all for making environment variables usable as config values (via default), as with the memcache options, but we should try to be consistent :)

jdfreder:
Thanks @bollwyvl !  Hopefully the API quota exhaustion is over now so the tests pass.

bollwyvl:
Hm, ten trials of restarting travis hasn't prevented all the `400`s from occurring for a single platform. My fear is we might have reached the end of that troubleshooting approach, and need to seriously consider how to move the test data into a fixture blob, either in another repo or a branch.

jdfreder:
It's easy enough to fork and branch the gist, I can do it tomorrow morning when I'm on the computer (on cell now).

jdfreder:
So I tried all combinations of,  
- forking @minrk 's gist  
- update the notebooks to v4  
- updating the notebooks to v4 and removing the blank cells  

and none seemed to affect the failure state of the tests.  Suggestions?  

(For the implementation, see c0d7ab6cf235ac38451073b10ee255f138de05df which I've now removed from this PR)

jdfreder:
... just a reminder, the tests that are failing are failing in master #531 and as far as I can tell are unrelated to the changes made by this PR.

bollwyvl:
I agree: these test fails are not related to the feature in question.

I don't think the tests I added create any additional requests, so I'll
have to assume the current master (production) would also fail the existing
batter... will verify.

minrk:
Tests should be fixed by #539.

minrk:
Needs rebase after #539 since server_cmd is now gone, in favor of server_args. I think it may need some fiddling to get the env test to work, since the tornado options are computed once before that test starts.

minrk:
I think you can clear the options with `tornado.options.options._options.clear()`, if you need to.

bollwyvl:
Awesome work, really, in this and the other PRs. I'll take a gander this
evening. Might have a bit of a rebase-off, but looking forward to it.

Thanks again!

jdfreder:
Hmm the last commit, should, hopefully, clear up the env test.  I'm not running the tests locally, so we'll wait to see what T says.

jdfreder:
Success, this is ready to go.

jdfreder:
@bollwyvl or @minrk want to take another look or merge?

minrk:
Thanks, @jdfreder!

PR #536, opened 2015-11-11
Fix recent 503 errors

### LINKED ISSUE #534: 503 backend read error
Recently, I am getting a lot of these (seem to involve a particular node):
screen

minrk:
The root cause is actually tornadoweb/tornado#1582, where failure to construct requests could eventually prevent future requests from ever being made. On our side, this was caused by failure to quote GitHub paths, which resulted in UnicodeEncodeErrors in PyCurl. Once 10 requests had been made to non-ascii GitHub URLs, no subsequent requests would ever be made.

While digging into this, I reworked how we wait for concurrent requests by using Futures, and fixed a few bugs in our logging of failed requests.

closes #534

### DIFF
--- a/nbviewer/log.py
+++ b/nbviewer/log.py
@@ -49,6 +49,6 @@ def log_request(handler):
         msg = msg + ' user-agent={agent}'
     if status >= 500 and status != 502:
         # log all headers if it caused an error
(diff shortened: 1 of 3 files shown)


minrk:
Yay, tests pass.

bollwyvl:
LGTM :shipit:

PR #538, opened 2015-11-11
make render_timeout an option

minrk:
and lower the default to 15 seconds.

It's actually this timeout that's causing the 503 errors on fastly, which has the same 20 second timeout, meaning that our slow renders never show the Working... page because fastly gives up immediately before the reply comes in.


jdfreder:
Thanks!

PR #539, opened 2015-11-11
Testing fixes

minrk:
- slides export can't handle empty notebooks, so use a notebook with a single markdown cell.
- run nbviewer test instance in a thread instead of a subprocess, which means we will get nice log capturing when tests fail, which makes fixing issues like the above a lot easier.


jdfreder:
Hallelujah!

PR #541, opened 2015-11-12
squashing deprecation warnings

bollwyvl:
Had to rebuild my nbviewer environment, and started seeing a bunch of deprecation warnings on run.

These may indeed break the 2.7 build, but we'll see what travis says :)

### DIFF
--- a/nbviewer/utils.py
+++ b/nbviewer/utils.py
@@ -7,11 +7,11 @@
 
 # https://docs.python.org/3.1/library/base64.html#base64.decodestring
 try:
-    from base64 import encodebytes as encodestring
-    from base64 import decodebytes as decodestring
+    from base64 import encodebytes
+    from base64 import decodebytes
 except ImportError:
-    from base64 import encodestring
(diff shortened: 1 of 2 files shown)


nbviewer/utils.py
```diff
@@ -5,7 +5,14 @@
 #  the file COPYING, distributed as part of this software.
 #-----------------------------------------------------------------------------
 
-import base64
+# https://docs.python.org/3.1/library/base64.html#base64.decodestring
+try:
+    from base64 import encodebytes as encodestring
+    from base64 import decodebytes as decodestring
+except ImportError:
+    from base64 import encodestring
```
minrk: probably makes more sense to import `deprecated as new` than `new as deprecated`, and update below.

bollwyvl:
Updated that. Back to test roulette!

I'm also trying to get my docker setup running again; always good to have a look at that :)

rgbkrk:
:smile:

bollwyvl:
destroy all spurious output!

PR #512, opened 2015-10-06
Slide templates

bollwyvl:
A resubmission of #511, but against the current pypi nbconvert (not master), to fix #506.

Adds a template path for nbconvert, the first use of which is used for a custom slides_reveal which just creates the body. This lets us get rid of the awful regex hack.

This is forked off of #508, whoops.

PR #544, opened 2015-11-25
bring slides up-to-date with nbconvert 4.1

### LINKED ISSUE #543: Slides mode doesn't work anymore on nbviewer
Dear all,

Many thanks for all your work on Jupyter, notebooks and nbviewer ! They are brilliant tools !

I've discovered this few days that all my notebooks located here : 
https://github.com/jhillairet/Fusion/tree/master/Python_Crash_Course/Slides

doesn't work anymore in with slide" mode viewer of nbviewer. The notebooks seems OK in normal mode, for e.g. :
http://nbviewer.ipython.org/github/jhillairet/Fusion/blob/master/Python_Crash_Course/Slides/5-Fusion_data.ipynb

but the "slide" mode, there is only one page showing up (can't navigate into the various slides), for e.g. : 
http://nbviewer.ipython.org/format/slides/github/jhillairet/Fusion/blob/master/Python_Crash_Course/Slides/5-Fusion_data.ipynb#/

I haven't changed anything since few weeks ago. Is it normal ? 

Best regards,

Julien

bollwyvl:
This fixes #543 by bringing basically reverting to the middle commit of #512. Wasn't paying attention to the nbconvert release cycle, where a bunch of the slide stuff got fixed.

I've checked this out in Docker, looks solid.

### DIFF
--- a/nbviewer/templates/nbconvert/slides_reveal.tpl
+++ b/nbviewer/templates/nbconvert/slides_reveal.tpl
@@ -1,46 +1,42 @@
 {%- extends 'basic.tpl' -%}
 
+
 {%- block any_cell scoped -%}
-{%- if cell.metadata.slide_type in ['slide'] -%}
-    <section>
-    <section>
-    {{ super() }}
-{%- elif cell.metadata.slide_type in ['subslide'] -%}
-    <section>
-    {{ super() }}
-{%- elif cell.metadata.slide_type in ['-'] -%}
-    {%- if cell.metadata.frag_helper in ['fragment_end'] -%}
-        <div class="fragment" data-fragment-index="{{ cell.metadata.frag_number }}">
-        {{ super() }}
-        </div>
-    {%- else -%}
-        {{ super() }}
-    {%- endif -%}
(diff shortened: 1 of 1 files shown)


bollwyvl:
Rethinking this, another approach would be to use a different template name, and import or extend the base template. This would prevent/reduce this kind of version mismatch bug in the future.

damianavila:
>Rethinking this, another approach would be to use a different template name, and import or extend the base template. This would prevent/reduce this kind of version mismatch bug in the future.

I agree.

bollwyvl:
>> use a different template name, and import or extend the base template. This would prevent/reduce this kind of version mismatch bug in the future.

> I agree.

Turns out this is non-trivial. The problem arises in how the `reveal_slides.tpl`'s `body` block is built. We basically want the  `super()` from `full.tpl`, not reveal. There is no way to get `super.super()` or something like that.

Given that this will take another upstream change anyway, I'm just going to garden the tests and merge this, and we can look at it fresh.

damianavila:
>Given that this will take another upstream change anyway, I'm just going to garden the tests and merge this, and we can look at it fresh.

That's ok. Thanks!

PR #565, opened 2016-01-21
install nbformat from master in Dockerfile

minrk:
to get reduced validation errors, which clog up nbviewer logs (~70% of logs now are validation errors)


willingc:
LGTM

PR #568, opened 2016-02-05
Support rendering of .snb files

vidma:
- [x] fix rendering
- [x] fix github fetching
- [x] configurable list of supported extentions

I'm sure guys from https://github.com/andypetrella/spark-notebook and https://github.com/Bridgewater/scala-notebook will be extremely excited.
At the moment their notebook format is 100% compatible with `ipynb`.

this tiny fix already renders fine github urls:
- lists available files, and links to .snb notebooks become active/clickable/renderable `http://docker:8080/github/andypetrella/spark-notebook/blob/master/notebooks`
- renders snb notebook: `http://docker:8080/github/andypetrella/spark-notebook/blob/master/notebooks/Flow%20Example.snb`

screen shot 2016-02-05 at 12 41 37

### DIFF
--- a/nbviewer/providers/base.py
+++ b/nbviewer/providers/base.py
@@ -56,7 +56,7 @@ except ImportError:
 
 date_fmt = "%a, %d %b %Y %H:%M:%S UTC"
 format_prefix = "/format/"
-
+notebook_file_extensions = ['.ipynb', '.snb']
 
 class BaseHandler(web.RequestHandler):
     """Base Handler class with common utilities"""
@@ -175,6 +175,13 @@ class BaseHandler(web.RequestHandler):
                 'name' : name,
             })
         return breadcrumbs
+
+    def is_notebook_file(self, path):
+        """return if a path is a notebook file
+
+        It checks for presence of file extension, such as '.ipynb'.
+        """
+        return any(path.endswith(ext) for ext in notebook_file_extensions)
 
     def get_page_links(self, response):
         """return prev_url, next_url for pagination
--- a/nbviewer/providers/github/handlers.py
+++ b/nbviewer/providers/github/handlers.py
@@ -167,7 +167,7 @@ class GitHubTreeHandler(GithubClientMixin, BaseHandler):
                 e['url'] = quote(e['url'])
                 e['class'] = 'fa-folder-open'
                 dirs.append(e)
-            elif file['name'].endswith('.ipynb'):
+            elif self.is_notebook_file(file['name']):
                 e['url'] = u'/github/{user}/{repo}/blob/{ref}/{path}'.format(
                 user=user, repo=repo, ref=ref, path=file['path']
                 )
@@ -256,7 +256,7 @@ class GitHubBlobHandler(GithubClientMixin, RenderingHandler):
             # filedata will be unicode
             filedata = contents
 
-        if path.endswith('.ipynb'):
+        if self.is_notebook_file(path):
             dir_path = path.rsplit('/', 1)[0]
             base_url = "/github/{user}/{repo}/tree/{ref}".format(
                 user=user, repo=repo, ref=ref,
--- a/nbviewer/providers/url/handlers.py
+++ b/nbviewer/providers/url/handlers.py
@@ -49,7 +49,7 @@ class URLHandler(RenderingHandler):
 
         if query:
             remote_url = remote_url + '?' + query
-        if not url.endswith('.ipynb'):
+        if not self.is_notebook_file(url):
             # this is how we handle relative links (files/ URLs) in notebooks
             # if it's not a .ipynb URL and it is a link from a notebook,
             # redirect to the original URL rather than trying to render it as a notebook


vidma:
P.S. Jupyther guys, your tests are broken for `py2.7` :(
`ImportError: cannot import name yaml2`

bollwyvl:
Thanks for the work!

This is a larger decision than nbviewer... I.e. i see no reason not to
merge this, (haven't reviewed technically) but the extension is one of the
things that ties the ecosystem together. Thus i see no problem in doing it,
but rather in enabling it on n.j.o.

I imagine you find hardcoded ipynb stuff across many of the projects. We'd
want to make sure this was done across the board, if we're going to do it.

We haven't gone so far as to register a mime type, but that would be the
"correct" answer... we're not going to get everyone to host with that, much
less utf-8 json correctly.

vidma:
ready to :rocket: 

--

P.S. it would be nice, but it doesn't make much sense to add testcase for `.snb` right now, [as it seems all `nbviewer`  tests relies on files from external repos](https://github.com/jupyter/nbviewer/blob/8747af0d59b875063aa524a929275bb62007428f/nbviewer/providers/github/tests/test_github.py#L21)

Carreau:
Hey !

Thanks for the pull-request ! As nick said, there I don't see any technical reason not to do that, and the request to have different extensions handled comes every now and then (not only by nbviewer, but other part of the tool chain). So far our answer has been no, for various reasons, 
but I don't reason not to restart. 

I think that if we do that, we have to be clear in communication that we support multiple extensions, and which one, and which is the boundary for new extensions. We decided in the beginning not to make `j<something>` for julia notebook. And in the case of `.snb` it can also be Samsung Phones Note files. So we have to be careful about that too.

You seem to have extra metadata. We would have to make sure our side of tools knows how to upgrade/downgrade the files correctly.

I'm excited if we can get the 2 project to get closer and interoperate. 

I'll ping @fperez and @ellisonbg who might have some more comment on the direction the project is taking, and wether it is worth investigating now, or a bit later are there are some refactoring in progress.

bollwyvl:
Thanks for the concise explanation, @Carreau!


Another option would be to have a URL param that would tell the backend:

> yep, i know this doesn't end in ipynb, but I'm telling you, it's a notebook

I'm thinking like `?notebook`. Or `?ipynb`. Or `?nb`. But we can't be going and parsing every JSON file out there, basically... there are perfectly valid reasons for nbviewer to proxy through a JSON file, and not try to render it as html.

> but other part of the tool chain

Since nobody will be linking into these files from an automated source (i.e. GitHub knows nothing of `.snb`):
https://github.com/andypetrella/spark-notebook/blob/master/notebooks/Flow%20Example.snb

Anyone who wants to plug in can just ask for it.

> You seem to have extra metadata

There is no metadata. There is only metadata. What I'm saying is, metadata demands only that it be an object, and have the kernel stuff.

I have wondered why we don't put an actual `schema` property in the ipynbs to make it self-describing, vs the major/minor stuff. But i digress.

minrk:
> **At the moment** their notebook format is 100% compatible with ipynb.

This seems like the crux of the problem.

If spark-notebook wants to fork the document format and assign a new file extension, that's AOK, but one of the things you lose by doing that is support of Jupyter notebook tools, such as nbviewer. A forked format that happens to be the same right now doesn't seem to me like something nbviewer should support, because it gives the illusion that we would be supporting future divergent changes to .snb, which we won't. Or are the spark-notebook folks planning to guarantee that .snb files will always be valid .ipynb going forward?

vidma:
> Or are the spark-notebook folks planning to guarantee that .snb files will always be valid .ipynb going forward?

we'll discuss this with other spark-notebook maintainers.
IMHO there is no issues currently (except that `jupyter` would not (always) be able execute `.snb`, e.g. because of some missing UI specific scala functions, like `display(variable)`).

a related topic, do you know how precisely works the automatic rendering of `ipynb` files in Github?
this could be  another motivator to collaborate for closely.

Carreau:
> we'll discuss this with other spark-notebook maintainers.

We will definitively iterate on notebook fileformat. We are just really conservative, and want to accumulate changes through time to only to (rare) major revision, and we make sure to have upgrade/downgrade mechanism. 

So if the scala team have some idea, they should be free to come forward. 

Note: we should have better track of potential enhancement for future versions.

> IMHO there is no issues currently (except that jupyter would not (always) be able execute .snb, e.g. because of some missing UI specific scala functions, like display(variable)).

Can you elaborate ? I'm curious of what the scala notebook have for that. 


> a related topic, do you know how precisely works the automatic rendering of ipynb files in Github?
> this could be another motivator to collaborate for closely.

Precisely maybe not, but from what I have gather: it uses nbconvert (as a library) with custom configuration for the conversion process.

minrk:
We do want to add 'importers' to nbconvert for things that can be converted to notebooks (e.g. ipymd, sage worksheets), and allow nbconvert to turn those into notebooks. When that's in place, .snb files will be the easiest possible case, since they can have an importer that's a no-op while the format continues to be compatible, and can grow to support any changes the snb folks choose to make down the line.

If .snb really do plan to only add metadata to the existing .ipynb format, then I'd be okay with this solution for now, and then move to an 'importer' when that's in place. The main bit that I don't like about this is that it's stating that .snb is actually a Jupyter notebook, which it either is (and should be .ipynb) or isn't (and shouldn't be treated as one).

fperez:
I don't like *at all* the idea of having one file type (Jupyter notebooks) growing multiple extensions like this.  If other projects are really using our format on-disk, then they should just call it an `ipynb`.  I pretty much concur with @minrk above, but I want to explicitly record that I don't think we should simply extend the extension list like this. We have over 50 kernels for Jupyter, if we accept this PR, we're basically saying to every one of those communities that we will take whatever extension they want.

While file extensions are a very poor way of representing format information, they are a real-world compromise that is extremely widely used.  Opening the door for myriad extensions that all point to the same file format will only make things hard for downstream tool users, who would have to know what this ever-growing list of extensions is to make their tools compliant.  This is very different from the case of image formats like JPEG or TIFF, that at most use a few variations on the base (`jpg`, `jpeg` and uppercase versions thereof).

Now, it may be that we should really revisit our decision to keep the `ipynb` name for our extension, if the "whiff of IPython" is what's bothering these other communities so much.  If that's the case, maybe we really need to rename our extension to say `jnb` or `jpynb` or something along those lines.  But in that case, tools would only have to support a new official extension plus `ipynb` for legacy reasons, not an open-ended list.

In summary, very strong -1 on this ever going in as-is.

Help us understand if the issue is the `ipy` in that name, and if so, we can consider a different solution.  But an open-ended extension list for our format is out of the question (@minrk's converter idea notwithstanding, which is a good one but solves a different problem, as it would allow these other hypothetical formats to evolve independently of ours).

rgbkrk:
I've totally been naming notebooks like `learning-rx.js.ipynb` and `data-munge.sc.ipynb`, especially when there are several in the same directory. Most importantly, I get to keep the same opener for the `.ipynb` extension while still showing to myself that it's a JS notebook or a Scala notebook at the terminal.

fperez:
Yup, I'd also thought of the `.lang.ipynb` option, but forgot it when writing the above, thanks for bringing that up @rgbkrk! I don't think we should try to make that automatic: it would require storing the short form of the language extension in the metadata for it to be done consistently, and it opens a few usability issues.  But as a user-driven solution, I think it's perfectly fine.  And if it catches on enough (perhaps combined with a change to a more neutral `jnb` or similar extension), we could revisit whether/how to implement it officially.

vidma:
yeah, sounds fair to close this 👍

PR #572, opened 2016-02-16
Cache upstream responses

### LINKED ISSUE #564: better caching
Traffic has increased to the point where we are hitting the GitHub API rate limit every couple of days. The main way that I see for us to deal with this is to cache our GitHub API requests more aggressively.

Right now, our caching strategy is very basic. We either cache a complete rendered result or nothing at all. We don't cache the results of any intermediate GitHub API requests.

I think there are a few things we can do to mitigate some of these things:

1. introduce caching of GitHub API request results
2. increase the baseline cache time from 10 minutes to 1 hour
3. increase cache duration when traffic is high
4. allow fallback on older cached values we wouldn't normally serve when rate limits are being hit

One alternative is to proactively aim for raw GitHub URLs when we can guess them (e.g. just for notebooks), skipping API requests in some cases, but that seems like being a bad API citizen, to me.

minrk:
For servers that support If-None-Matching, If-Modified-Since cache headers, lean on upstream 304 responses for cache hit detection.

If an upstream request gives 304, the cached response is re-used.

Cached responses are saved for a hardcoded two hours, for now. Since we only support 304-based cache detection, a long expiry should not be a problem.

Since the GitHub API uses these headers and 304 responses don't count against your rate limit, this should significantly improve our rate limit consumption.

I also changed the error message when GitHub requests are failing due to the rate limit, so instead of a 400 with opaque "The error was: 403", it is a 503 error and explicitly mentions the GitHub rate limit and come back later.

closes #564

### DIFF
--- a/nbviewer/app.py
+++ b/nbviewer/app.py
@@ -42,9 +42,9 @@ from .providers import default_providers, default_rewrites
 from .providers.local import LocalFileHandler
 
 try:
-    from .providers.url.client import LoggingCurlAsyncHTTPClient as HTTPClientClass
+    from .providers.url.client import NBViewerCurlAsyncHTTPClient as HTTPClientClass
 except ImportError:
-    from .providers.url.client import LoggingSimpleAsyncHTTPClient as HTTPClientClass
+    from .providers.url.client import NBViewerSimpleAsyncHTTPClient as HTTPClientClass
 
 
 from .log import log_request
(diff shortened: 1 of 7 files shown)


rgbkrk:
Great, this is really forward looking for users.

minrk:
Deployed.

bollwyvl:
Awesome! Go, go nbviewer! :horse_racing:

PR #575, opened 2016-02-21
Add support for a 'script' format

yuvipanda:
- Runs the ipynb file through the ScriptExporter
- Sets Content-Type to text/plain
- Allows formatters that need to specify Content-Type


yuvipanda:
Avenue for (later?) improvement - add support for 'raw' formats that don't go through the jinja templating system.

minrk:
What would be the point of a 'raw' format, since all nbviewer does is render existing notebooks already publicly available? If there's no change to make, isn't that just downloading directly from the original URL?

yuvipanda:
Oh, 'raw' as in 'does not go through jinja2' - right now I unnecessarily
pass the output straight through Jinja for no reason. Same for a PDF
format...

yuvipanda:
Thanks for merging!

yuvipanda:
@minrk and a 'raw' as you mentioned would still be useful for local providers.

minrk:
Ah, raw as in doesn't go through jinja makes sense (though PDF *does* go through jinja to latex, and then to PDF).

yuvipanda:
Yup, just no second layer of Jinja :)

PR #576, opened 2016-02-21
remove unneeded dependency on notebook package

minrk:
- fetch notebook source only for LESS build step
- depend directly on nbconvert, IPython


rgbkrk:
:+1:

Issue #585, opened 2016-03-17
"Open with Binder" button


Carreau:
Just discuss that with @freeman-lab and @andrewosh , the new version of binder should be able to be queried for "has this notebook/url a binder".

I propose that when we render a notebook that can be started a binder, show binder button next to the "view on GitHub" button.

Thoughts ?

rgbkrk:
Sounds good to me.

parente:
@Carreau would the same approach you took with Open with Binder work here for any github repo not already setup for Binder?

Carreau:
> @Carreau would the same approach you took with Open with Binder work here for any github repo not already setup for Binder?

I'm not sure I understand. For "Open with Binder" I just assume any github repository works. Like it's just reg-ex based :-P

PR #586, opened 2016-03-18
[wip] binder button

bollwyvl:
Hooray #585! :tada: I have wanted to see this since the original announcement, great news @Carreau (and, of course, @freeman-lab, @andrewosh, @rgbkrk). Had to throw something down quick! Screenshots at the bottom!

### What is done?
This is the simplest jinja/tornado stuff I could think of to add a binder button to the github provider. Since (at least in my current understanding) a binder either exists or not for a whole _repo_, not an individual notebook, I'm showing it on both the notebook page and the repo tree list.

I've added some minimal failing tests... instead of never showing the button (since we can't hit an API yet), it just always shows the button. The failing test is for a project that probably won't have a binder... namely binder itself.

### What might we do/discuss next?
- [ ] pick icons
- [ ] need to know the API endpoint, once it's a available
- [ ] cached, async client request to determine whether there is binder or not
  - [ ] will we need an API key or something?
- [ ] in the case of a notebook, can we deep-link directly to it?
  - otherwise the user will have to navigate first in nbviewer, then again in the repo... which might be obstructed by the automatic index.ipynb loading, etc.

### What else could be done?
Presumably one could also show something on the user's repo listing page, i.e. a run-button-per-project. Quick snapshot of a user's reproducibility!

### What's it look like?
No big surprise. If there's a better icon, or we actually want to use the binder logo, that's great.

screen shot 2016-03-18 at 12 26 36 am

screen shot 2016-03-18 at 12 28 01 am

Hooray!

### DIFF
--- a/nbviewer/providers/base.py
+++ b/nbviewer/providers/base.py
@@ -506,9 +506,11 @@ class RenderingHandler(BaseHandler):
                 app_log.info("failed to test %s: %s", self.request.uri, name)
 
     @gen.coroutine
-    def finish_notebook(self, json_notebook, download_url, provider_url=None, 
+    def finish_notebook(self, json_notebook, download_url, provider_url=None,
                         provider_icon=None, provider_label=None, msg=None,
-                        breadcrumbs=None, public=False, format=None, request=None):
+                        breadcrumbs=None, public=False, format=None,
+                        runner_icon=None, runner_label=None, runner_url=None,
+                        request=None):
         """render a notebook from its JSON body.
 
         download_url is required, provider_url is not.
@@ -550,6 +552,9 @@ class RenderingHandler(BaseHandler):
             provider_url=provider_url,
             provider_label=provider_label,
             provider_icon=provider_icon,
+            runner_url=runner_url,
+            runner_label=runner_label,
+            runner_icon=runner_icon,
             format=self.format,
             default_format=self.default_format,
             format_prefix=format_prefix,
--- a/nbviewer/providers/github/handlers.py
+++ b/nbviewer/providers/github/handlers.py
@@ -35,8 +35,11 @@ from .client import AsyncGitHubClient
 PROVIDER_CTX = {
     'provider_label': 'GitHub',
     'provider_icon': 'github',
+    'runner_icon': 'play-circle-o',
+    'runner_label': 'Binder'
 }
 
+BINDER_TMPL = u"http://mybinder.org/repo/{user}/{repo}"
 
 class GithubClientMixin(object):
     @property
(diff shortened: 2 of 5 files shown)


freeman-lab:
@bollwyvl super cool to see this! A few quick thoughts

- overall layout seems great, I'm all for a simple "play" style button
- for now yes, there's one binder per repo
- but we can deep link to particular notebooks, so it should be easy to have the play button link to either the top-level folder (which will then default to index.ipynb binder-side), or the specific notebook you're viewing
- the simple check for whether a binder exists or not won't require an API token (other things will)

@andrewosh can jump in as well with further thoughts!

bollwyvl:
Thanks for the feedback!

> - deep link to particular notebooks

Ah, had to dig into the source to find it: here's a link for posterity

```
http://mybinder.org/repo/K3D-tools/K3D-jupyter/examples/line.ipynb#anchor
```

> - the simple check for whether a binder exists

I have tried:
```
http://api.mybinder.org/apps/{user}/{repo}/status
```

This is a bit slow (couple of seconds). Is there a better URL to use? Or is this there a better on coming in the new version? We cache the heck out of everything, so unfortunately we can't really, say, treat the first check as a throwaway, with the hope that the boolean will be true the next time, as it won't even hit the rendering mechanism. Then we might be racing the cache.

Been tinkering with this, but putting this stuff directly in the github provider is making me a little uneasy.

I propose we introduce a new high-level extension point, `Runner`, that takes the final rendered context in `finish_notebook`, and has a chance to change it. `BinderRunner` would be the first one, but there's no telling what people might uncover given the appropriate interface. And documentation.
 
This will allow:
- multiple runners (probably still stack them in config, first hit wins)
- self-hosters to disable binder integration altogether

The runner itself should have a couple things, and ideally an implementer should be able to configure just a few parts and get something interesting:
- `run_url` is a high-level coroutine that will eventually return a runnable URL or `None`. the default implementation would use...
  - `check` a normal method which determines (cheaply!) if a notebook is even something the runner would want
    - `BinderRunner` would be mainly concerned with the `provider_url`... though may be interested in the formatter as well... i.e. launching right into slides, etc. there is also a potential cheap check if there's actually a markdown/URL in the body of the notebook that matches the binder button for the current notebook
  - `verify` more expensively, saying definitively whether a context is runnable, returning a verification object (i.e. with transformed information that might be useful for the url) or `None`. again, if the link actually appears in the notebook, (and matches the repo) we're good otherwise...
    - i.e. check `https://api.mibinder.org/apps/foo/bar/status`
  - `_run_url` actually performs the transformation
    - pretty straightforward, actually, in this case!

Carreau:
This has been inactive for a few month I'm going to close for now and tag with a closed-pr tag to be easily found.

Carreau:
Thanks all !

Issue #587, opened 2016-03-24
Github Enterprise 400 Bad Request


Drunkar:
Hi, thank you for great useful library.

When I used nbviewer docker container for Github Enterprise, I got 400 Bad Request.
It seems to be redirected to the login page.
Does anyone can run and render *.ipynb with Github Enterprise?

Thanks.

#### docker run command:

```
docker run --name nbviewer -p 8080:8080 -e 'GITHUB_API_TOKEN=<USER_NAME>:<ACCESS_TOKEN>' -e 'GITHUB_API_URL=http://<GITHUB_ENTERPRISE_URL>/api/v3/' jupyter/nbviewer
```

#### error log:

```
[E 160324 22:16:14 base:525] Failed to render file from url: http://**** (My repository URL)
    Traceback (most recent call last):
      File "/opt/conda/lib/python3.5/site-packages/nbformat/reader.py", line 14, in parse_json
        nb_dict = json.loads(s, **kwargs)
      File "/opt/conda/lib/python3.5/json/__init__.py", line 319, in loads
        return _default_decoder.decode(s)
      File "/opt/conda/lib/python3.5/json/decoder.py", line 339, in decode
        obj, end = self.raw_decode(s, idx=_w(s, 0).end())
      File "/opt/conda/lib/python3.5/json/decoder.py", line 357, in raw_decode
        raise JSONDecodeError("Expecting value", s, err.value) from None
    json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
    
    During handling of the above exception, another exception occurred:
    
    Traceback (most recent call last):
      File "/home/mine/nbviewer/nbviewer/providers/base.py", line 523, in finish_notebook
        nb = reads(json_notebook, current_nbformat)
      File "/opt/conda/lib/python3.5/site-packages/nbformat/__init__.py", line 74, in reads
        nb = reader.reads(s, **kwargs)
      File "/opt/conda/lib/python3.5/site-packages/nbformat/reader.py", line 58, in reads
        nb_dict = parse_json(s, **kwargs)
      File "/opt/conda/lib/python3.5/site-packages/nbformat/reader.py", line 17, in parse_json
        raise NotJSONError(("Notebook does not appear to be JSON: %r" % s)[:77] + "...")
    nbformat.reader.NotJSONError: Notebook does not appear to be JSON: '<!DOCTYPE html>\n<html lang="en" class=...
```

willingc:
[...]

PR #589, opened 2016-03-26
Add rewrite rules for Github Enterprise.

Drunkar:
PR for issue #587 .


rgbkrk:
Great! Would you be able to add a test so that a future contributor (including myself) doesn't break the implementation you're relying on for GHE (especially since we won't see it in production ourselves)?

Drunkar:
Thank you, @rgbkrk .
I think it is very important for repository's health, too!
I'm writing the test.

Drunkar:
Oops, something got wrong with git.
I close this PR once and create new one.

PR #591, opened 2016-04-01
Delete the gitter badge

rgbkrk:
Another room we don't use much that ends up getting diverted to jupyter/jupyter.


Carreau:
Can you replace by a link to jupyter/help ?

PR #595, opened 2016-04-03
relax gist hash length regex

minrk:
Allow >= 20 characters, rather than strictly 20 because gist IDs are now 32 characters, and there's no reason to be strict about it that I can see.

### DIFF
--- a/nbviewer/providers/gist/handlers.py
+++ b/nbviewer/providers/gist/handlers.py
@@ -186,10 +186,10 @@ def default_handlers(handlers=[]):
     """Tornado handlers"""
 
     return handlers + [
-        (r'/gist/([^\/]+/)?([0-9]+|[0-9a-f]{20})', GistHandler),
-        (r'/gist/([^\/]+/)?([0-9]+|[0-9a-f]{20})/(?:files/)?(.*)', GistHandler),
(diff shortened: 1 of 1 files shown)


nbviewer/providers/gist/handlers.py
```diff
@@ -186,10 +186,10 @@ def default_handlers(handlers=[]):
     """Tornado handlers"""
 
     return handlers + [
-        (r'/gist/([^\/]+/)?([0-9]+|[0-9a-f]{20})', GistHandler),
-        (r'/gist/([^\/]+/)?([0-9]+|[0-9a-f]{20})/(?:files/)?(.*)', GistHandler),
-        (r'/([0-9]+|[0-9a-f]{20})', GistRedirectHandler),
-        (r'/([0-9]+|[0-9a-f]{20})/(.*)', GistRedirectHandler),
+        (r'/gist/([^\/]+/)?([0-9]+|[0-9a-f]{20,})', GistHandler),
```
bollwyvl: Right, until GH starts using a [different encoding](https://en.bitcoin.it/wiki/Base58Check_encoding#Base58_symbol_chart)... but much closer to Once And For All...

bollwyvl:
Can't find any other instances of weirdness with gist ids. Hooray! :sunglasses:

PR #597, opened 2016-04-05
Send metrics via statsd

### LINKED ISSUE #596: Add statsd metrics
Similar in vein to https://github.com/jupyter/jupyterhub/issues/459. I think having graphs and metrics about nbviewer usage would be awesome :D  http://grafana.wikimedia.org/ is example of similar dashboard from Wikimedia.

And yes, I am volunteering to do this (unless someone objects)

yuvipanda:
Fixes #596

### DIFF
--- a/nbviewer/app.py
+++ b/nbviewer/app.py
@@ -217,6 +217,9 @@ def make_app():
         localfile_path=os.path.abspath(options.localfiles),
         fetch_kwargs=fetch_kwargs,
         mathjax_url=options.mathjax_url,
+        statsd_host=options.statsd_host,
+        statsd_port=options.statsd_port,
+        statsd_prefix=options.statsd_prefix
     )
 
     # handle handlers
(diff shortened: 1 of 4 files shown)


yuvipanda:
I'm going to add some more.

yuvipanda:
I think this is good to go now! It collects timing and success/fail info for renders, and splits them per-renderer cleanly. We can add more later if we want.

Now, who do I ping to help get this deployed on nbviewer.jupyter.org? :) @rgbkrk?

rgbkrk:
:heart: the timing of renders

rgbkrk:
Something I didn't ask - is statsd going to do any blocking network actions? Are there examples or even code that connects tornado up nicely with statsd? Sorry I didn't have this question before. The sheer nature of traffic to nbviewer made me think about it.

rgbkrk:
Added you on to our ops handbook org @yuvipanda.

yuvipanda:
@rgbkrk So it's sending out an UDP packet (via https://docs.python.org/3/library/socket.html#socket.socket.sendto) and that technically blocks but is fast enough to not matter at these sizes I think (since it isn't TCP). If we run into issues we can always write a simple tornado based implementation of the statsd client protocol (which is pretty simple anyway...). 

If you think this would be a problem (I think it should be ok), I'm happy to do that first.

rgbkrk:
We can always try it and see. After all, we'd now be measuring the impact of that too I assume.

Wonder what @minrk thinks.

minrk:
We just need a statsd timer to measure the statsd calls...

We should measure it, but I would be surprised if it impacts anything, since it is a UDP publish of a tiny message, as mentioned.

rgbkrk:
> We just need a statsd timer to measure the statsd calls...

ha.

PR #605, opened 2016-04-25
set blockquote font-size to inherit

### LINKED ISSUE #604: Blockquote font too large
The font size for `<blockquote>` elements is awkwardly large:

```css
blockquote {
    font-size: 17.5px;
}
```

I didn't find this in the repo and I don't know where it comes from, but it seems to end up in http://nbviewer.jupyter.org/static/build/styles.css.

In the live Notebook as well as in the HTML-output of `nbconvert`, the font size is the same as the surrounding text, which looks decent:

```css
blockquote {
    font-size: inherit;
}
```

Could you please change this?

minrk:
rather than bootstrap default of 1.25 x base

closes #604

### DIFF
--- a/nbviewer/static/less/variables.less
+++ b/nbviewer/static/less/variables.less
@@ -13,6 +13,8 @@
 
 @breadcrumb-bg: #fafafa;
 
+@blockquote-font-size: inherit;
+
 @footer-bg: #979797;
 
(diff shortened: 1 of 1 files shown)


willingc:
Thanks @minrk for closing @mgeier's issue #604

Issue #616, opened 2016-06-10
docker build fails at `invoke less`


ChristianKniep:
Hey there,

when I build your image I get the following error:

```
Step 18 : RUN invoke less
 ---> Running in 717bb57270f8
Traceback (most recent call last):
  File "/usr/local/lib/python3.4/dist-packages/invoke/tasks.py", line 141, in argspec
    context_arg = arg_names.pop(0)
IndexError: pop from empty list

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/bin/invoke", line 11, in <module>
    sys.exit(program.run())
  File "/usr/local/lib/python3.4/dist-packages/invoke/program.py", line 269, in run
    self._parse(argv)
  File "/usr/local/lib/python3.4/dist-packages/invoke/program.py", line 325, in _parse
    self.load_collection()
  File "/usr/local/lib/python3.4/dist-packages/invoke/program.py", line 473, in load_collection
    coll = loader.load(coll_name) if coll_name else loader.load()
  File "/usr/local/lib/python3.4/dist-packages/invoke/loader.py", line 53, in load
    module = imp.load_module(name, fd, path, desc)
  File "/usr/lib/python3.4/imp.py", line 235, in load_module
    return load_source(name, filename, file)
  File "/usr/lib/python3.4/imp.py", line 171, in load_source
    module = methods.load()
  File "<frozen importlib._bootstrap>", line 1220, in load
  File "<frozen importlib._bootstrap>", line 1200, in _load_unlocked
  File "<frozen importlib._bootstrap>", line 1129, in _exec
  File "<frozen importlib._bootstrap>", line 1471, in exec_module
  File "<frozen importlib._bootstrap>", line 321, in _call_with_frames_removed
  File "/srv/nbviewer/tasks.py", line 29, in <module>
    @invoke.task
  File "/usr/local/lib/python3.4/dist-packages/invoke/tasks.py", line 267, in task
    return Task(args[0], **kwargs)
  File "/usr/local/lib/python3.4/dist-packages/invoke/tasks.py", line 58, in __init__
    self.positional = self.fill_implicit_positionals(positional)
  File "/usr/local/lib/python3.4/dist-packages/invoke/tasks.py", line 149, in fill_implicit_positionals
    args, spec_dict = self.argspec(self.body)
  File "/usr/local/lib/python3.4/dist-packages/invoke/tasks.py", line 144, in argspec
[...]

PR #617, opened 2016-06-10
updating to invoke 0.13.0

bollwyvl:
Thanks to @ChristianKniep on #616, looks like `invoke` changed its API since last we tried to build stuff. This breaks the docker image, and of course local installation.

This is a quick fix, adding the `ctx` argument to each of the tasks, which is now the preferred way to call other tasks: i.e. `ctx.run` vs `invoke.run`.

Since travis has raised it's game significantly with respect to docker, I'm adding an explicit test of our docker stuff there to assist in testing against docker for future work. Another approach would be to have the dockerfile itself run the tests... really just a oneliner.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,9 @@ install:
 - pip install -r requirements.txt
 - invoke bower
 - invoke less
-script: invoke test
+script:
+  - invoke test
+  - docker-compose build
 env:
   global:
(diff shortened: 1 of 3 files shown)


bollwyvl:
> Since travis has raised it's game significantly with respect to docker, I'm adding an explicit test of our docker stuff there to assist in testing against docker for future work. Another approach would be to have the dockerfile itself run the tests... really just a oneliner.

Would change the run environment a good bit, so removed this.

rgbkrk:
Thanks!

bollwyvl:
Thanks @rgbkrk for the super-quick review!

ChristianKniep:
Slendid! I'll give it a try on Monday...

Issue #621, opened 2016-08-15
Github Enterprise: Top-right button "View on Github" links to https://github.com instead of Github Enterprise URL


juanrg:
Hi!

There is a little bug in the github enterprise version of the github provider. The little "View on github" button is hardcoded, so it will always link directly to github even when setting GITHUB_API_URL. For clarity, I am talking about this one:

![nbviewer_github](https://cloud.githubusercontent.com/assets/2692663/17670248/4a3b5488-6311-11e6-8fb0-37f5eadc5a97.png)

I have a solution by adding another system variable, GITHUB_URL. It could be extracted from GITHUB_API_URL, but it is probably safer to keep them separate. Can send a pull request if this solution is fine with you.

parente:
A second env var, or parsing of the API env var, seems as good a fix as any for this issue.

vaguiar:
@parente Wanted to check if this is still and issue? https://github.com/jupyter/nbviewer/pull/623 seems to provide a fix. 

I checked with @ivanov at the PyCon19 sprints and he suggested asking for more clarification.

parente:
@vaguiar I missed the referenced PR when doing a bulk triage of all the nbviewer issues. I think it's safe to close until/unless someone with GitHub enterprise can say there's still an issue.

Thanks for checking. Good luck sprinting!

PR #623, opened 2016-08-29
Fixed a bug on the Github Enterprise version of the github provider

juanrg:
By default the behaviour stays the same. Now added an environment variable GITHUB_URL, in a similar fashion to the variable GITHUB_API_URL already in place by a previous patch. See #621

### DIFF
--- a/nbviewer/providers/github/handlers.py
+++ b/nbviewer/providers/github/handlers.py
@@ -39,6 +39,9 @@ PROVIDER_CTX = {
 }
 
 
+def _github_url():
+    return os.environ.get('GITHUB_URL') if os.environ.get('GITHUB_URL', '') else "https://github.com/"
+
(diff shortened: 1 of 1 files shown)


nbviewer/providers/github/handlers.py
```diff
@@ -100,7 +103,8 @@ def get(self, user):
                 url=repo['name'],
                 name=repo['name'],
             ))
-        provider_url = u"https://github.com/{user}".format(user=user)
+  
+        provider_url = u"{github_url}{user}".format(user=user, github_url = _github_url())
```
rgbkrk: This should likely form the path either with a utility function or as:

```
u"{github_url}/{user}".format(user=user, github_url = _github_url())
```

Normalization ftw.

PR #624, opened 2016-09-03
Change hardcoded url to a call to pip

christophelec:
I replaced the harcoded call to pypi.org with the use of pip. This allows to use the configuration of pip transparently, in case there is an internal mirror to use, and simplify the import

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -5,21 +5,16 @@ from __future__ import print_function
 
 import os
 import hashlib
-import json
 import shutil
 import tempfile
 import sys
-import tarfile
-
-try:
-    from urllib.request import urlretrieve
-except ImportError:
-    from urllib import urlretrieve
+from zipfile import ZipFile
+import pip
 
 import invoke
 
 NOTEBOOK_VERSION = '4.1.0' # the notebook version whose LESS we will use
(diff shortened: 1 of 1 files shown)


rgbkrk:
:cool: It does look like this currently isn't working right now though on Travis. Does pip need an upgrade?

christophelec:
I used '.' as the folder in which to download, I think this is what causes the issue.

I'll change to APPROOT and try on my side.

christophelec:
Ok, I found the issue : for some reason on debian, pip favours zip instead of tar gz.

I tried to look around, but it does not seem possible to specify tar.gz instead of zip in pip, only binary or source (tar.gz or zip) package.

So I simply changed to use zip instead, as it is possible to handle it as native python anyway.

tasks.py
```diff
@@ -5,21 +5,16 @@
...
-except ImportError:
-    from urllib import urlretrieve
+from zipfile import ZipFile
+import pip
 
 import invoke
 
 NOTEBOOK_VERSION = '4.1.0' # the notebook version whose LESS we will use
-NOTEBOOK_CHECKSUM = 'b597437ba33538221008e21fea71cd01eda9da1515ca3963d7c74e44f4b03d90' # sha256 checksum of notebook tarball
+NOTEBOOK_CHECKSUM = '24cac518544caaeca34b39931cbc1a4ac580078f2044ba09adb9dee01a72dc84' # sha256 checksum of notebook tarball
```
rgbkrk: -> ...of notebook zipfile

PR #632, opened 2016-10-03
update link to new asset

CamDavidsonPilon:
We updated the urls in Bayesian Methods for Hackers (new PyMC3 version is available), so this PR changes the urls on the frontpage too. 

cc @Carreau for review

### DIFF
--- a/nbviewer/frontpage.json
+++ b/nbviewer/frontpage.json
@@ -34,7 +34,7 @@
       },
       {
         "text": "Probabilistic Programming",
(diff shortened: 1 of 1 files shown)


bollwyvl:
Great, thanks for the update!

Hm, this still links to PyMC2... did you want to link it to PyMC3?
> github/CamDavidsonPilon/Probabilistic-Programming-and-Bayesian-Methods-for-Hackers/blob/master/Chapter1_Introduction/Ch1_Introduction_PyMC3.ipynb

CamDavidsonPilon:
hm, part of me wants to keep PyMC2 for historical reasons, but then again, out with the old and in with the new! I'll update now to PyMC3 👌

Carreau:
Thanks !

CamDavidsonPilon:
Hey, was this deployed to production? I'm still seeing a 404 on http://nbviewer.jupyter.org/

PR #636, opened 2016-10-26
Support --base-url

parente:
Make nbviewer relocatable behind a proxy (e.g., as a Jupyter Hub service)

### DIFF
--- a/nbviewer/app.py
+++ b/nbviewer/app.py
@@ -48,7 +48,7 @@ except ImportError:
 
 
 from .log import log_request
-from .utils import git_info, jupyter_info
+from .utils import git_info, jupyter_info, url_path_join
 
 #-----------------------------------------------------------------------------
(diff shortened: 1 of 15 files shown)


parente:
The approach of prefixing all URLs created across all providers and core pages is getting too invasive. I'm going to back up and try a `from_base` function for use in templates that only affects absolute paths and nothing else.

parente:
New approach is cleaner, but the slide down of the header is not working. Need to look.

parente:
@minrk We spoke about this work briefly in gitter last week. It's prereq to #639 which will make it possible to protect nbviewer with Jupyter Hub service auth.

I think this PR can be reviewed independently of the auth implementation. It's useful for relocating nbviewer in certain situations regardless.

minrk:
Great, thanks @parente!

PR #639, opened 2016-10-31
Support running as a Jupyter Hub service

parente:
Depends on #636 for base-url support. Reads the JUPYTERHUB_SERVICE_PREFIX, JUPYTERHUB_API_URL, and JUPYTERHUB_API_TOKEN env vars. If present, uses them to authenticate user access to the viewer.

Useful particularly when using the localfile provider.

This is a work in progress. Feedback is welcome. Things to do / consider:

* [x] Add doc about the new support and parameters.
* [-] ~~Should the code optionally import jupyterhub.auth and use the HubAuth implementation there, with its caching mechanism, rather than reimplementing a tinier version here?~~ Let's start with the simple implementation inline here and switch to the more feature-filled one later if warranted.


minrk:
Cool! I merged #636, if you want to rebase to get a smaller diff.

You might also want to consider looking at JUPYTERHUB_SERVICE_URL for the ip/port to bind to ([example](https://github.com/jupyterhub/jupyterhub/blob/master/examples/service-whoami/whoami.py#L33))

parente:
Getting back to this now. I'll rebase and continue adding support for more of the env vars. I'd like add redirection to `JUPYTERHUB_BASE_URL/hub/login` instead of 401 errors as well.

parente:
It dawns on me that the JupyterHub proxy URL needs to be exposed as an environment variable so that a service can redirect a browser to the Hub login page.

parente:
@minrk @willingc When you have a few minutes, I think this one is ready for review.

README.md
```diff
@@ -184,3 +184,26 @@ Formats are ways to present notebooks to the user.
 If you'd like to write a new format, open a ticket, or speak up on [gitter][]!
 We have some work yet to do to support your next big thing in notebook
 publishing, and we'd love to here from you.
+
+## Securing the Notebook Viewer
+
+You can run the viewer as a JupyterHub 0.7+ service. Doing so prevents users who have not authenticated with the Hub from viewing any notebooks. This setup can be useful for protecting access to local notebooks rendered with the `--localfiles` option.
```
willingc: Let's go ahead and link to the JupyterHub services section: http://jupyterhub.readthedocs.io/en/latest/services.html

Also:
~~Doing so~~``Running the viewer as a service prevents users who have not authenticated with the Hub...``

parente:
Thanks @willingc. I made the doc edits.

minrk:
> It dawns on me that the JupyterHub proxy URL needs to be exposed as an environment variable so that a service can redirect a browser to the Hub login page.

We could use tornado's own login_url and get_current_user settings for this, which would handle the checking and redirecting for us. Doing so would require putting `@web.authenticated` on everything, but it would mean redirects and URLs would be correct.

parente:
>  Doing so would require putting @web.authenticated on everything, but it would mean redirects and URLs would be correct.

I think I solved the redirect issue with @willingc 's help with the current `prepare` implementation. (I tried hard to avoid adding @web.authenticated everywhere: surely I'd miss something.)

parente:
I rebased this on master to fix the merge conflict. I think it's OK for now to stick with the `prepare` implementation which should redirect correctly based on `JUPYTERHUB_BASE_URL`.

parente:
🎈

Issue #640, opened 2016-11-01
localfile provider fails when a base_url is provided


ptorrestr:
Hi,

When `base_url` is provided in the command line, the localfile provider will fail. For instance, suppose this argument line:

```
python3 -m nbviewer --port=8080 --localfiles=/some/dir --base_url=/path
```
You will not be able to access  `localhost:8080/path/localfile/some/file` since the url handler for this provider does not include `base_url`.

PR #641, opened 2016-11-01
Fix for #640

ptorrestr:
Include base_url for localfile provider.


parente:
Good catch! I totally missed that case because the handler is optional. 🍰

PR #642, opened 2016-11-07
Add directory support to --localfiles

parente:
Allows navigation of directories under the localfiles root. Renders them in a page layout similar to the treelist template.


parente:
Thanks @minrk

PR #650, opened 2016-12-13
Add jupyter-js-widgets embedder

SylvainCorlay:
This will complete inline rendering in nbconvert once https://github.com/jupyter/nbconvert/pull/482 gets in a release.

Ping @michaelpacer

### DIFF
--- a/nbviewer/templates/layout.html
+++ b/nbviewer/templates/layout.html
@@ -157,6 +157,7 @@
     </footer>
   {% endblock footer %}
 
+  <script src="https://unpkg.com/jupyter-js-widgets@2.0.*/dist/embed.js"></script>
   <script src="{{ static_url("components/bootstrap/js/bootstrap.min.js") }}"></script>
   <script src="{{ static_url("components/headroom.js/dist/headroom.min.js") }}"></script>
   <script src="{{ static_url("components/headroom.js/dist/jQuery.headroom.min.js") }}"></script>


michaelpacer:
My instinct is that this should be in nbconvert not nbviewer if nbconvert
is going to support widgets at all.

I just would prefer to have a handle that can autogenerate the link without
making ipywidgets a strict dependency for nbconvert (so that it picks up
whatever is the most recent included in the released version of
ipywidgets). Could this be surfaced in the notebook? Then we can hard code
it for now, but expect to drop in a variable once it's available.

michaelpacer:
"If nbconvert is going to support widgets at all" being somewhat of a
foregone conclusion right now.

However, I think that nbviewer is not on the beta of nbconvert and I don't
want to release until we fix the header ids, which got pushed back because
widgets. So I think it still makes sense to merge this here. Just keeping
in mind a future change to remove it once it's in nbconvert as a
dynamically defined link (with a fallback to this link hardcoded if the
ipywidgets either isn't on the system or if the ipywidgets too old to
surface this link appropriately).

That makes me think this can be merged now, as is.

SylvainCorlay:
> My instinct is that this should be in nbconvert not nbviewer if nbconvert
is going to support widgets at all.

Yes, I don't know. That is why I did not want to include it in the first nbconvert PR anyway.

> However, I think that nbviewer is not on the beta of nbconvert and I don't
want to release until we fix the header ids, which got pushed back because
widgets. So I think it still makes sense to merge this here. Just keeping
in mind a future change to remove it once it's in nbconvert as a
dynamically defined link (with a fallback to this link hardcoded if the
ipywidgets either isn't on the system or if the ipywidgets too old to
surface this link appropriately).

Could we make a patch release of nbconvert with the new widget thing? That would allow us to iterate quickly on the potential issues with the embedder...

PR #651, opened 2016-12-14
Add jupyter-js-widgets embedder

SylvainCorlay:
@minrk

### DIFF
--- a/nbviewer/templates/layout.html
+++ b/nbviewer/templates/layout.html
@@ -157,6 +157,7 @@
     </footer>
   {% endblock footer %}
 
+  <script src="https://unpkg.com/jupyter-js-widgets@2.0.*/dist/embed.js"></script>
   <script src="{{ static_url("components/bootstrap/js/bootstrap.min.js") }}"></script>
   <script src="{{ static_url("components/headroom.js/dist/headroom.min.js") }}"></script>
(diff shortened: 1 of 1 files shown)


bollwyvl:
Hooray! So much ❤️ to @SylvainCorlay and @michaelpacer for the push to make this happen.

This won't be the last mimetype we want to expose here.

Just to stick up for folk that host this in-house, we should probably at least make this a configurable option, that, at worst, just was a list of external bundles at unpkg, but at best was a mechanism that let you add all the bundles of interest to the page... though better still would be being able to pick from the whitelist by looking at the ipynb and only loading the appropriate ones (if it ends up taking much of a performance hit to load a lot of them).

Hooray!

SylvainCorlay:
> Just to stick up for folk that host this in-house, we should probably at least make this a configurable option, that, at worst, just was a list of external bundles at unpkg, but at best was a mechanism that let you add all the bundles of interest to the page... though better still would be being able to pick from the whitelist by looking at the ipynb and only loading the appropriate ones (if it ends up taking much of a performance hit to load a lot of them).

I agree with that. I am definitely going to propose an embedder that also does not attempt to connect to cdns to render custom widgets, for people use cases where you want to limit what may be loaded on the page.

Madhu94:
@bollwyvl I opened a separate issue for this here https://github.com/jupyter-widgets/ipywidgets/issues/1894

PR #659, opened 2016-12-31
Fix view as code icon

parente:


minrk:
nbviewer shouldn't require jupyter_client anywhere. Where does that come from?

minrk:
Nevermind, I see it's a bug in nbconvert 5.0, fixed here: https://github.com/jupyter/nbconvert/pull/499 . We can remove when nbconvert gets another release.

parente:
Thanks for the multiple merges, @minrk.

minrk:
Came back from the holiday with a big pile of nice PR 🎁 s. Happy 2017!

PR #667, opened 2017-02-12
display time from now instead of UTC time

randy3k:
screen shot 2017-02-11 at 10 10 58 pm

Show time from now instead of UTC time in directory view of local files.


parente:
Oh wow. I didn't realize nbviewer was already including moment. 👍

randy3k:
It is also used in the [footbar](https://github.com/jupyter/nbviewer/blob/262016ec2059746550ff5cf86b0e243bb84e88a6/nbviewer/templates/layout.html#L182) 😃 

screen shot 2017-02-11 at 10 25 30 pm

PR #668, opened 2017-02-12
api_token is not needed for managed service

randy3k:
instead, we need to change the working directory to where nbviewer is located.

### DIFF
--- a/README.md
+++ b/README.md
@@ -198,8 +198,8 @@ c.JupyterHub.services = [
         'name': 'nbviewer',
         # the interface and port nbviewer will use
         'url': 'http://127.0.0.1:9000',
-        # the api token for nbviewer to check user auth
-        'api_token': 'my-secret-nbviewer-token',
+        # the path to nbviewer repo
+        'cwd': '<path to repo>',
         # command to start the nbviewer
         'command': ['python', '-m', 'nbviewer']
     }


parente:
The changes suggested here change the nbviewer example from a auth-required, external Hub service (e.g., running in the provided Docker container) to an open-access (i.e., no auth required) managed Hub service. Instead of swapping the one for the other, how about we include both with text describing the two examples?

randy3k:
The changes that I made reflect what was written on README, which is "managed service". README also states that

> You can also run the nbviewer instance as an externally managed JupyterHub service, but must set the requisite environment variables yourself.

I guess we should add another section for it.

parente:
You're right about the example being about a managed service. That's what I get for trusting my memory and only reading the diff without the context.

I've re-read that README section in full now. The title of that section is currently *Securing the Notebook Viewer* and the first paragraph states:

> Running the viewer as a service prevents users who have not authenticated with the Hub from acccessing the nbviewer instance. This setup can be useful for protecting access to local notebooks rendered with the --localfiles option.

I suggest leaving the `api_token` in that section since the token is what enables the auth described. If you'd like to contribute another section about running nbviewer without auth either as a managed or unmanaged service, that'd be great.

Also, note that the `cwd` is only necessary if nbviewer is being used from a git clone instead of from an  installed copied of the package.

randy3k:
The api_token is only required for externally managed service. It is not needed in our case. 

You are right about the cwd argument. However, there is no mentioning on how to install nbviewer as a package in README though.

minrk:
@parente all services get API tokens. The managed service doesn't need it in config because it is generated at launch time. External services need it because it must be generated and passed to both the service and the Hub to get the same value.

parente:
Sorry for the confusion!

PR #681, opened 2017-03-07
Add "Show/Hide Inputs" button to navbar

parente:
Adds a button to toggle cell inputs to the navbar. Tries to collapse whitespace between visibly empty cells by removing padding and border while avoiding too much jitter and excessive animation. Super basic but very effective.

toggle-inputs

The original idea is from @csaid (http://chris-said.io/2016/02/13/how-to-make-polished-jupyter-presentations-with-optional-code-visibility/). I tweaked it a bit in http://mindtrove.info/code-hiding-on-nbviewer/. Any reason this shouldn't live in nbviewer itself?

### DIFF
--- a/nbviewer/static/less/notebook.less
+++ b/nbviewer/static/less/notebook.less
@@ -11,7 +11,6 @@
   text-align: center;
 }
 
-
 @media (max-width: 767px) {
   div.input, div.output_area {
     box-orient: vertical;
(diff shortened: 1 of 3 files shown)


mariusvniekerk:
Nice

parente:
Also, to raise up attribution from the commit message:

* Original idea from @csaid at http://chris-said.io/2016/02/13/how-to-make-polished-jupyter-presentations-with-optional-code-visibility/

Carreau:
cc @michaelpacer who is working on tags. 

I'm going to _guess_ that sometime you don't want to hide all the code cells . I have no objections.

parente:
Agreed, @Carreau. I think this can grow over time to support tags for shown/hidden cells, the [dashboard layout metadata](http://jupyter-dashboards-layout.readthedocs.io/en/latest/metadata.html) , and other info stored in the notebook about how to present it. I've seen users get good mileage with the snippets linked above, and figured they would be a good starting point.

parente:
Correcting a "at" above: /cc @mpacer

mpacer:
Is there a reason you did this with js and not pure css animations (ala https://github.com/mpacer/hiding_tags_nbconvert/blob/master/toggle_hidden.tpl)? 

also, is there a reason to not have individual I hide buttons on the margin in addition to the global button? It could be hidden until hover which would harm discoverability but would surface exactly the feature you want by default (by instantly toggling the state for all code 
cells).

parente:
> Is there a reason you did this with js and not pure css animations (ala https://github.com/mpacer/hiding_tags_nbconvert/blob/master/toggle_hidden.tpl)?

Only that I had working JS code to submit. I'm in favor of using a pure CSS approach. I'll look at updating this PR.

> also, is there a reason to not have individual I hide buttons on the margin in addition to the global button?

Same reason as above mainly. Plus I'm a fan of building things incrementally with user feedback.

mpacer:
The biggest advantage that JS could give over my pure CSS approach that still (almost entirely) uses the pure CSS approach would be to have the JS extract and then fix the height of the cells (rather than leaving them on `auto`). Because then you won't need to use the `max-height` trick that I used to make animations occur (since CSS animations can't be applied to an `auto` defined height since they can't inspect the object directly).

Regardless, I think that it would be cleaner if you toggled a class and just had altered properties associated with the class in the actual CSS files (rather than a raw encoding).

One of the advantages of that is that once it's based on just toggling a class, the button for toggling/untoggling individual cells is really straightforward to implement (especially since we have a working example of it).

parente:
Thanks @mpacer for the pointers. Here's the new rendition using a CSS class a just a tiny bit of JS to toggle it in the nav bar.

toggle-inputs-css

There's no nbconvert config in nbviewer at the moment, other than one for the slideshow resources. We can follow-up by adding one to get the per-cell checkboxes in.

nbviewer/templates/layout.html
```diff
@@ -19,6 +19,16 @@
 {%- endmacro %}
 
 
+{% macro head_js_icon(js, name, icon, accesskey=None) -%}
+  <li>
+    <a href="javascript:{{js}}" title="{{name}}" {%if accesskey %}accesskey={{accesskey}}{% endif %}>
+      <span class="fa fa-{{icon}} fa-2x menu-icon"></span>
+      <span class="menu-text">{{name}}</span>
```
mpacer: I like this, but it might be worth expanding the totality of the js that's included…

maybe something like (note this won't work and may not be valid js

```javascript
<script>
for (x in $.('hide_input_cells')){
   $(x.id).height($(x.id).height())
}
</script>
```

Or should this go below since you're actually passing the javascript _in_ there and there is minimal js in this actual macro…

NB: I've not seen this use of an `href` or `a` before, I'm intrigued.

nbviewer/templates/notebook.html
```diff
@@ -14,6 +13,8 @@
     {% endif %}
   {% endfor %}
 
+  {{ layout.head_js_icon("$('body').toggleClass('hide_input_cells');", "Show/Hide Inputs", "eye", "i") }}
```
mpacer: I think this is where the above code might actually go (inside the first argument (currently `"$('body').toggleClass('hide_input_cells');"`))…

I actually looked this up a bit more and think these are most of the necessary details though there are still things that need to be checked… E.g., If you can do this without explicitly iterating over all the elements that need to be toggled. 
I'll look at this tomorrow after sleeping on it. 

```javascript
for (x in $.('.input_area')){//input areas are the only things that have the class input area
   x.height(x.height()); //first call sets it, second call retrieves it, they may need to be separated
// the key insight is that is that x.height() should return the pixel height and x.height(val) sets the height to that val.
   x.toggleClass('hide_input_cell');//applies the toggle
}
```

Right now, I think you're applying 'hide_input_cell' to the entire body, but I figure this would be better to apply to the individual input areas. It also makes it a lot easier to open the way to have checkboxes in the future if can toggle individual cells. Of course, that means that the logic I wrote above won't work for the hide/show all button because then they can be toggled individually and this will just swap state. That's probably going to need more than one line though.

nbviewer/static/less/notebook.less
```diff
@@ -20,8 +19,60 @@
...
+    transition: max-height var(--in-time) var(--transition-path-in), padding .0s step-end;
+    -webkit-transition: max-height var(--in-time) var(--transition-path-in), padding .0s step-end;
+    -moz-transition: max-height var(--in-time) var(--transition-path-in), padding .0s step-end;
+    -o-transition: max-height var(--in-time) var(--transition-path-in), padding .0s step-end;
+    max-height: var(--max-height-big);
+}
+
+body.hide_input_cells div.input {
+    overflow:hidden;
+    max-height: var(--max-height-small);
```
parente: I can give height a shot.

parente: I see the same shadowing when animating the `height` by setting the `div.input` height on load and adding `height: 0px !important` when `.hide_input_cells` is applied.

toggle-inputs-css

mpacer:
Is it possible that this is an optical illusion? Or are the pixels actually
doing that shadowing?

Because that should have worked as far as I can understand what is
happening.

If it has to do with line height… one possibility is to first squash all
text into one line. Then you have a common place to be working from. But I
think a simple drag/squash overlay would look awkward. It wouldn't look
good with fonts of different sizes and styles. I don't know if it would
look good for even one font. But it (or something like it) would solve the
irregularity by reducing everything to a common case.

I'll read more about css animations this week and get back to you. It
should be possible to make the animation smooth and continuous for any size.

parente:
Sorry for not responding sooner.

> Is it possible that this is an optical illusion? Or are the pixels actually doing that shadowing?

I believe it's shadowing. I can see the entire contents of the cell input appear over all of the content instantly, while the space to accommodate it all grows.

Since we're heading back toward using more JS to fix these issues, does it make sense to go back to the original JS implementation?

parente:
Returning to this PR, do folks think we try to work out the shadowing issues with CSS, go with a straight JS approach, or abandon it altogether for now and revisit when there's a final nbformat spec that we can code toward?

damianavila:
> revisit when there's a final nbformat spec that we can code toward?

I think it would follow that option.

parente:
OK. Closing this out.

PR #694, opened 2017-05-04
mathjax 2.7.1, notebook 5.0.0

minrk:
will get new table styles from notebook 5.0


parente:
I pulled the branch, ran the dev build locally, and tested an instance against a handful of notebooks. The new table styling shows up fine and MathJax appears to be working. I didn't spot any regressions.

Thanks @minrk!

PR #709, opened 2017-07-11
prepare to put it on pypi

### LINKED ISSUE #704: release jupyter/nbviewer as a pypi package
It'd be so awesome if we can just `pip install -U nbviewer` and then `python -m nbviewer`.

minrk:
call it 1.0 and update a couple of links in setup.py

closes #704

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,0 +1,2 @@
+[wheel]
+universal = True
--- a/setup.py
+++ b/setup.py
@@ -29,6 +29,7 @@ pkg_data = {
     )
 }
 
+
 setup_args = dict(
     name = "nbviewer",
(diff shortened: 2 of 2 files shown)


parente:
Does the setup.py trigger the invoke logic which bundling of the jupyter notebook frontend assets that also have to wind up in the package?

minrk:
@parente I just noticed that it doesn't, so you'd have to be careful to run it before upload. I'll add that before finishing this PR.

parente:
LGTM. If you're done, I can give it a whirl here if you want before merging.

parente:
I've been remiss in testing it locally. Let's merge and address problems in follow-on PRs.

PR #710, opened 2017-07-24
support customized style

pltchuong:


nbviewer/formats.py
```diff
@@ -45,7 +45,7 @@ def default_formats():
             'nbconvert_template': 'slides_reveal',
             'label': 'Slides',
             'icon': 'gift',
-            'test': lambda nb, json: '"slideshow"' in json,
+            'test': lambda nb, json: 'celltoolbar' in nb.metadata and nb.metadata['celltoolbar'] == 'Slideshow',
```
parente: Does `'celltoolbar' in nb.metadata` only evaluate to true if the notebook author had the slide toolbar visible when they last saved the notebook? Is that the intended behavior change here?

pltchuong: Yes that's true. The reason is, at first we have our notebook as slide, then we disable it by change all cell to - (there is no other option), but with the original condition it always catch our notebook as slide. Also, there is always a case we have text slideshow in the json itself

parente:
Returning to this many many moons later ...

I think we can split out the custom stylesheet change and merge that in without any changes. The slideshow detection need some work I think. If we merge that part of the PR as-is, anyone who had turned off the slideshow toolbar and was relying on the public nbviewer to offer the gift box icon to View as Slides will be surprised when it no longer appears.

We could scan all cell metadata looking for the presence of at least one `slide_type != -` value. I suspect this test will take longer than the current json-string scan for the token `slideshow`, but it won't be prone to false positives, retains the current behavior, and should have an unnoticeable impact on the total time it takes to parse the notebook file with nbformat and render it with nbconvert.

PR #716, opened 2017-09-05
Allow customization of frontpage text

dbdean:
Following on from my earlier pull request (https://github.com/jupyter/nbviewer/pull/683), I've added top-level 'title', 'subtitle', 'text' and 'show_input' fields to the frontpage.json file, with the original sections under a top-level 'sections' tag.

Current-style frontpage.json files with only the list of sections will still work as they did before, but may be considered deprecated with this PR. I have updated the test cases to have a new schema that validates the new frontpage.json style, and the tests pass.

This has allowed me to fully customize the text in the front page for my organisation without having to maintain a separate fork, although I may need to do so when I want to change the styling (one thing at a time!).

I have remove the --no-frontpage-input option that I introduced in https://github.com/jupyter/nbviewer/pull/683, as it is no longer needed, and I am assuming it was just me using it. If needed, it could be kept with a deprecation warning.

I'm happy to modify this pull request as needed to get it into the main repository.

Thanks!


dbdean:
Any news on this PR? I'd be happy to help if any adjustments might be required to get it into master.

parente:
Sorry. I have not taken the time to look at any of the nbviewer PRs. I'll try to get to them in the near future if someone else doesn't get here first.

dbdean:
Any news? I'd love to help get this into the repo. If there are any blockers, please let me know if I can do anything to help remove them.

nbviewer/templates/index.html
```diff
@@ -1,10 +1,15 @@
 {% extends "layout.html" %}
 
 {% block body %}
-  {% if not no_frontpage_input %}
```
parente: The origin `no_frontpage_input` flag also hid the jumbotron header. Does the page look OK with that element still in the DOM when there's no title, subtitle, or text?

dbdean: Let me check on that, and get back to you.

dbdean:
Looking back, I intended that the title field would be required, so hiding the jumbotron wasn't an issue. However, given that you *could* hide it before, I've added another commit here to make it no longer required, and to hide the jumbotron itself if nothing is being shown in it.

Thanks for your help getting this done!

nbviewer/templates/index.html
dbdean: I've added a new commit to hide the jumbotron if nothing is contained within it.

parente:
Looks good now. Thanks for the feature @dbdean.

PR #732, opened 2017-10-14
Upgrade to notebook 5.2.0

gnestor:


parente:
I think that’s all it takes, besides redeploying to the live site. Thanks!

PR #733, opened 2017-10-15
Fixes link to JupyterHub services documentation

sethwoodworth:
The previous link 404'd as the documentation was reorganized to add a parent directory `reference`.


parente:
Thanks for the fix! 🍰

PR #738, opened 2017-11-07
Upgrade ipywidgets to v7, with a fallback for v6

### LINKED ISSUE #717: Jupyter widgets 7.0 update
https://github.com/jupyter/nbviewer/blob/5ac2643d687f4e78c5b8a4d50ead8255b2c65bb1/nbviewer/templates/layout.html#L160 is very specific to Jupyter widgets 6.x. It should be updated to support both widgets 6.x and 7.x (which would require checking the metadata and determining which version, and then either using that embed url or using the new embed URL (see http://ipywidgets.readthedocs.io/en/stable/migration_guides.html?highlight=migrate#updating-embedded-widgets)

scottdraves:
fix #717

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,7 @@ before_install:
 - npm install .
 
 install:
+- pip install --upgrade setuptools pip
 - pip install -r requirements.txt
 - invoke bower
 - invoke less
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,6 +4,7 @@ FROM debian:jessie
 MAINTAINER Project Jupyter <jupyter@googlegroups.com>
 
 ENV DEBIAN_FRONTEND=noninteractive
+ENV LANG=C.UTF-8
 RUN apt-get update \
   && apt-get install -y -q \
     build-essential \
--- a/nbviewer/app.py
+++ b/nbviewer/app.py
@@ -244,6 +244,7 @@ def make_app():
         hub_api_token=os.getenv('JUPYTERHUB_API_TOKEN'),
         hub_api_url=os.getenv('JUPYTERHUB_API_URL'),
         hub_base_url=os.getenv('JUPYTERHUB_BASE_URL'),
+        ipywidgets_base_url=options.ipywidgets_base_url,
     )
 
     if options.localfiles:
@@ -298,6 +299,7 @@ def init_options():
     define("statsd_port", default=8125, help="Port on which statsd is listening for metrics on statsd_host", type=int)
     define("statsd_prefix", default='nbviewer', help="Prefix to use for naming metrics sent to statsd", type=str)
     define("base_url", default='/', help='URL base for the server')
+    define("ipywidgets_base_url", default="https://unpkg.com/", help="URL base for ipywidgets package", type=str)
 
 
 def main(argv=None):
--- a/nbviewer/providers/base.py
+++ b/nbviewer/providers/base.py
@@ -194,6 +194,10 @@ class BaseHandler(web.RequestHandler):
         return self.settings['mathjax_url']
 
     @property
+    def ipywidgets_base_url(self):
+        return self.settings['ipywidgets_base_url']
+
+    @property
     def statsd(self):
         if hasattr(self, '_statsd'):
             return self._statsd
@@ -255,6 +259,7 @@ class BaseHandler(web.RequestHandler):
             "static_url": self.static_url,
             "from_base": self.from_base,
             "google_analytics_id": self.settings.get('google_analytics_id'),
+            "ipywidgets_base_url": self.ipywidgets_base_url,
         }
 
     def breadcrumbs(self, path, base_url):
--- a/nbviewer/providers/base.py
+++ b/nbviewer/providers/base.py
@@ -603,9 +603,10 @@ class RenderingHandler(BaseHandler):
                 app_log.info("failed to test %s: %s", self.request.uri, name)
 
     @gen.coroutine
-    def finish_notebook(self, json_notebook, download_url, provider_url=None, 
+    def finish_notebook(self, json_notebook, download_url, provider_url=None,
                         provider_icon=None, provider_label=None, msg=None,
-                        breadcrumbs=None, public=False, format=None, request=None):
+                        breadcrumbs=None, public=False, format=None, request=None,
+                        title=None):
         """render a notebook from its JSON body.
 
         download_url is required, provider_url is not.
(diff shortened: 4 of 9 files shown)


scottdraves:
@jasongrout

bollwyvl:
I'd love to see this make it in but do wonder about our ongoing ability to maintain it and/or extend to other mime renderers. If there's any way that we can make future modifications one-liners to a single (json) file, that would be preferable. 

At a minimum, I'd like to see the `unpkg` urls moved to something configurable...

jasongrout:
The code looks okay to me. @bollwyvl - do you have any tips on how to make that URL configurable?

mjurowicz-xsolve:
@jasongrout @bollwyvl any updates here?

jasongrout:
I *think* I figured out how to make it configurable. Basically, look in https://github.com/jupyter/nbviewer/blob/master/nbviewer/providers/base.py and https://github.com/jupyter/nbviewer/blob/master/nbviewer/app.py for "mathjax_url" and do something similar for the ipywidgets url (you'll probably have two ipywidgets urls, one for v6, one for v7).

jasongrout:
If we want to do this configuration in a separate PR, I'm okay with that as well. This PR is good and self-contained as an upgrade as-is, though I agree with @bollwyvl that having a configurable URL would be nice.

jasongrout:
In fact, if you've tested this PR as-is, I'd be inclined to merge this as-is and then we can do the configuration changes in a separate PR. Let me know if you want to do the configuration in this one, though.

jasongrout:
Actually, looking more into the source code for nbviewer, I do have one more comment on this PR. As-is, the widgets code is added in the layout.html template, but I think that's the wrong place (and it was wrong before - not your fault!). The layout template is used for things like the page listing all of the notebooks, etc. Instead the widgets code should be added to the https://github.com/jupyter/nbviewer/blob/master/nbviewer/templates/notebook.html template, right next to the mathjax block at https://github.com/jupyter/nbviewer/blob/ffda79bb4c74a370824e676ed23e0448ac778bcb/nbviewer/templates/notebook.html#L51. Then it will be loaded just for actual notebooks. Can you make this change in this PR?

And please let us know if you want to do the configuration changes in this PR, or if we should open an issue/PR for it separately.

bollwyvl:
Hi, sorry for the delay. It's absolutely good to go, as we want widgets :)

I guess there are two dimensions of configuration:
- what mimetypes do we support?
- where do we get them?

So the easiest answer from the downstream perspective would be that the mimetype itself would carry enough information to determine what we need to render it. Having to pre-parse the content that will get re-parsed again seems pretty rough... what if the json is bad? 

The we'd need a simple file someplace, probably realizable in the config file, or at least pointed to by the config file.

This could then either be requested with `$.get` and used directly, or, saving a step, generate some boilerplate.

Anyhow... I boiled all this down to some ideas here:

https://nbviewer.jupyter.org/gist/anonymous/a44bf8d688135d58588c07dd61ec90ee

bollwyvl:
> In fact, if you've tested this PR as-is,

Have not :blush:

> mathjax_url

Right, this pattern, but have a plan for something more general and replaceable. If the path of justice is `unpkg`, _and_ it's possible to manually dereference this locally to be hosted disconnected from the internet, huzzah! `unpkg` is by neccessity not very policed, (vs cdnjs, jsdelivr, etc) and in fact has been used as a C&C for exploits. I wouldn't be surprised if it isn't blacklisted in a few places already.

> loaded just for actual notebooks

Yes, this. Also, probably go ahead and wrap it in a `{% block renderers %}{% endblock %}`. At very least, a local admin could override the template and just kill that block.

jasongrout:
Note that jsdelivr also supports the 'anything that is an npm package is available' - we've looked at switching to it.

scottdraves:
thanks for the comments.
moving from layout.html to notebook.html should be no problem, that sounds better.
look for a commit on monday.
making it more configurable sounds like a new issue...

jasongrout:
> thanks for the comments.
> moving from layout.html to notebook.html should be no problem, that sounds better.
> look for a commit on monday.

Great, thanks!

> making it more configurable sounds like a new issue...

Sounds good! I made an issue: #739.

bollwyvl:
Had a chance to try this out locally: looking good!

The updates are looking robust, but more byzantine: I'd like us to start thinking about this in a more systematic way that will permit handling the anticipated explosion of [mime renderers](https://github.com/jupyterlab/jupyter-renderers/issues/3). Anyhow, out of scope for this issue, I suppose.

One regression I'm seeing is that (on chrome, but not FF), the imported material design stuff is changing some site chrome, initially have noticed link colors:
screen shot 2017-11-13 at 9 54 45 am

vs 6.x

screen shot 2017-11-13 at 9 55 25 am

This is from an eager un-namespaced `a` CSS selector with a freaky vendor selector:
screen shot 2017-11-13 at 9 56 35 am

We can either get more aggressive on this in the nbviewer selectors, but it might be a good thing to push upstream... not sure if it's worth holding up the

scottdraves:
This is a big blocker for us and i think the CSS fix lies elsewhere...

scottdraves:
ping...

jasongrout:
Thanks for the ping - it's in my queue to look at today.

jasongrout:
@bollwyvl - do you know where that `a` css styling is coming from? Is it coming from widgets?

jasongrout:
I reproduced the css styling problem, and tracked it down to the widgets embed script. I'll be solving that upstream.

I think this code looks good. I probably would have used `jupyterwidgets_base_url` or `npm_base_url`, but that can be a different PR.

jasongrout:
I also confirmed this works great for ipywidgets 6.

Thanks again!

scottdraves:
🐬 🐬

jasongrout:
The blue link issue is traced back to JupyterLab: https://github.com/jupyterlab/jupyterlab/issues/3257

jasongrout:
Great question. I don't know how deployment works for nbviewer.

jasongrout:
Let's ask on the jupyterhub gitter channel.

scottdraves:
looks like it happened.

jasongrout:
Great! And we're working on fixing that CSS issue upstream, which should show up when we push an updated jupyterlab and ipywidgets.

minrk:
This should be deployed to nbviewer. Sorry for taking so long. @jasongrout I've added you on the private nbviewer-deploy repo, so you should be able to do a deploy if you want.

jasongrout:
Thanks!

PR #744, opened 2017-12-07
implement the "--localfile_abspath" command line switch

MarkusTeufelberger:
See #743

### DIFF
--- a/nbviewer/app.py
+++ b/nbviewer/app.py
@@ -241,6 +241,7 @@ def make_app():
         gzip=True,
         render_timeout=options.render_timeout,
         localfile_path=os.path.abspath(options.localfiles),
+        localfile_use_abspath=options.localfile_abspath,
         fetch_kwargs=fetch_kwargs,
         mathjax_url=options.mathjax_url,
         rate_limiter=rate_limiter,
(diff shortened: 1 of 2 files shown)


nbviewer/app.py
```diff
@@ -281,6 +282,7 @@ def init_options():
     define("debug", default=False, help="run in debug mode", type=bool)
     define("no_cache", default=False, help="Do not cache results", type=bool)
     define("localfiles", default="", help="Allow to serve local files under /localfile/* this can be a security risk", type=str)
+    define("localfile_abspath", default=False, help="Use the absolute instead of the real path for local files, don't dereference symbolic links", type=bool)
```
parente: Minor suggestion: Maybe name the parameter `localfile_follow_symlinks`. Personally, I always need to google to remember which one of realpath or abspath dereferences the symlinks.

MarkusTeufelberger: Nice suggestion, I'm not great at naming stuff. ;-)

Will do.

nbviewer/providers/local/handlers.py
```diff
@@ -37,7 +37,10 @@ class LocalFileHandler(RenderingHandler):
 
     @property
     def localfile_path(self):
-        return os.path.realpath(self.settings.get('localfile_path', ''))
+        if self.settings.get('localfile_follow_symlinks'):
+            return os.path.abspath(self.settings.get('localfile_path', ''))
+        else:
+            return os.path.realpath(self.settings.get('localfile_path', ''))
```
parente: I think the logic here is inverted. `os.path.realpath` resolves (follows) symlinks to their targets (https://docs.python.org/3.6/library/os.path.html#os.path.realpath). `os.path.abspath` just gets the absolute path of the symlink itself. I think these need to be flipped, along with the code blocks below too.

MarkusTeufelberger: Ah, I see it now, that was an oversight on my end in the option renaming. Thanks for the review!

parente:
Thanks for the PRs @MarkusTeufelberger! 🍰

PR #771, opened 2018-05-30
Fixes for installation, pip 10

minrk:
- restore running invoke prior to build_py/sdist/develop,
  which avoids installation missing the built files
- fix invocation of `pip download` for pip 10 (`pip.main` is gone)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,13 +13,11 @@ before_install:
 - sudo apt-get install -qq libzmq3-dev pandoc libcurl4-gnutls-dev libmemcached-dev
 - pip install --upgrade setuptools pip
 - pip install -r requirements-dev.txt
-- npm install .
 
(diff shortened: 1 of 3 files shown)


parente:
Thanks @minrk . Further failures are due to GitHub rate limiting. Did we have a GH API key setup on Travis at some point that went missing or did I imagine that?

parente:
I added a personal GH token as env var to Travis with no oauth scope. Let's see if that gets the job done and isn't horribly insecure.

parente:
Nope. That's right. PR builds don't honor those env vars for security. Looks like @rgbkrk set up one or more encrypted env vars in the .travis.yml 4 years ago. I bet one of those was a GH key that was working until recently. I can have a look at updating them later if no one gets to it first.

PR #774, opened 2018-06-21
Fix ISO 8601 formatting for moment.js

parente:
We need to make UTC timestamps conform to true ISO-8601 by appending Z(ulu). Without a timezone, the spec says it should be treated as local time which is not what we want and causes moment.js on the frontend to show times in the past or future depending
on the user's timezone.

The code already assumes the server is providing UTC modification times by using datetime.utcfromtimestamp(). This PR just corrects the frontend rendering of that value.

https://en.wikipedia.org/wiki/ISO_8601#Time_zone_designators


parente:
Thanks @minrk

PR #777, opened 2018-06-30
Add issue templates

parente:
Adding templates for feature requests, bug reports, and general questions. Including guidance about what to do if the feature/bug/question is not about nbviewer but rather another Jupyter project.


parente:
In case you haven't seen it (I hadn't!) here's what the user sees after clicking *New Issue* when multiple issue templates exist:

screen shot 2018-06-30 at 4 57 13 pm

parente:
I've revised the PR to refer users to jupyter/help for general questions about use and installation and to not include a general question template.

I'm going to merge what I have as a way of giving users more guidance about when to open issues here vs elsewhere. We can revise the templates or remove them altogether if they don't pan out. I'd like these in place before starting to tidy up the backlog.

PR #785, opened 2018-07-27
Add support for the Content-Security-Policy header

parente:
Adds a command line `--content_security_policy=<value>` option which becomes a `Content-Security-Policy: <value>` HTTP header in response to all requests. Makes it a little less scary to open notebooks of unknown origin and content.

The default value of `connect-src 'none';` is totally open for discussion but felt like a reasonable safeguard to me for people who might run the server locally. (It disallows XHR, WebSocket, Fetch, etc. to any domain.)


minrk:
Makes sense

PR #794, opened 2018-09-03
Update Python, use multistage build

parente:
Switches the Dockerfile to use conda and a multistage build so that JS buildchain tools are kept out of the final image. The image size is still quite large because of pandoc and all the Python packages though. The main benefit here is the ability to update the version of Python.


Carreau:
Thanks !

parente:
FWIW, DockerHub reports the compressed image size is now 239 MB. It was roughly 355 MB if I remember correctly before this change.

Issue #828, opened 2019-05-09
Insufficient error logging in localfile provider


krishnatg:
If the nbviewer is started with the `--localfiles` option and there are incorrect access permissions given to the provided directory path, nbviewer throws a `404` error when the `/localfile` URL is hit. However, there are no additional error logs to help understand the gap. I had to resort to adding debug logs to pinpoint the problem, and figured it might make sense to get them added here.

parente:
Closed by #829. Thanks!

PR #829, opened 2019-05-09
Additional logging for file permission issues

krishnatg:
- Added warning logs for instances where the `local` provider encounters file permission issues (https://github.com/jupyter/nbviewer/issues/828)

### DIFF
--- a/nbviewer/providers/local/handlers.py
+++ b/nbviewer/providers/local/handlers.py
@@ -115,7 +115,7 @@ class LocalFileHandler(RenderingHandler):
             return False
 
         if not os.path.exists(fullpath):
-            app_log.warn("path: '%s' does not exist" % fullpath)
+            app_log.warn("path: '%s' does not exist", fullpath)
             return False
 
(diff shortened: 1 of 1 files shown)


nbviewer/providers/local/handlers.py
```diff
@@ -115,6 +115,7 @@ def can_show(self, path):
             return False
 
         if not os.path.exists(fullpath):
+            app_log.warn("path: '%s' does not exist" % fullpath)
```
parente: Thanks for the PR. I think adding all of these should be fine. Could you please change them all to use the standard Python logging module approach of passing arguments as the second parameter before we merge? 

For example:

`app_log.warn("path: '%s' does not exist", fullpath)`

should work, is standard, and saves the interpolation when the log level is set higher than the function.

parente:
@krishnatg Thanks again for the PR. 🍰

krishnatg:
Thank you for the accepting this PR 😁

PR #833, opened 2019-05-30
(1) Fixed FAQ page, (2) made front page compatible with JupyterHub

krinsman:
**(1)** Currently the Jinja2 in the `faq.md` file references a Python-Markdown extension, "HeaderId", which was removed from Python-Markdown in 2018; see this commit:

https://github.com/Python-Markdown/markdown/commit/11270135194922e0f5cfc739b69fe39f7337a0f9

Also, according to that commit, the HeaderId extension was deprecated and then removed from Python-Markdown because it can be replaced entirely by the `toc` extension, which is already referenced in the `faq.md` file.

Therefore, when trying to run nbviewer locally, with `faq.md`, and clicking on the FAQ link from the front page, one hits a "500 Internal Server Error" page, and the log files show that the error is
```ModuleNotFoundError: No module named 'headerid(level=3)'```.

After I removed the reference to `headerId` in `faq.md` and re-ran nbviewer locally, the FAQ page loaded correctly (no 500 Internal Server Error) and looked like the corresponding FAQ page on nbviewer.jupyter.org, which presumably is the intended behavior.

**(2)** With the way `frontpage.json` was currently written, when one clicked on any of the links on the front page, when nbviewer was hosted as a(n externally managed) service on JupyterHub, the link would take the user to, e.g. `baseURL/hub/services/gist/fonnesbeck/2352771` instead of the intended behavior `baseURL/services/nbviewer/gist/fonnesbeck/2352771`. In the former case we get a 404 error, whereas in the latter case nbviewer actually pulls up the intended files.

The root URL for nbviewer when hosted as a JupyterHub service is `baseURL/services/nbviewer`. In order to interact with nbviewer, the GET requests all have to be to URLs underneath this. So in particular in order to avoid 404ing and to actually interact with the nbviewer service instead of the hub itself, it's necessary for the URIs on the front page to send the user to URLs starting with `baseURL/services/nbviewer`. Any other prefix, e.g. `baseURL/hub/services/` causes 404 errors. The intended behavior doesn't happen without the beginning `/` in the URI. I don't know why, but URIs starting with the empty string have the incorrect prefix `baseURL/hub/services` prepended, while URIs starting with `/` have the correct prefix `baseURL/services/nbviewer` prepended.

**Comments:**
These are minor bugs and what is proposed is a minor fix -- hopefully it is still worth your consideration.

@parente I'm not sure who is the maintainer of this project right now.

### DIFF
--- a/nbviewer/frontpage.json
+++ b/nbviewer/frontpage.json
@@ -8,17 +8,17 @@
      "links":[
        {
          "text": "IPython",
-         "target": "github/ipython/ipython/blob/6.x/examples/IPython%20Kernel/Index.ipynb",
+         "target": "/github/ipython/ipython/blob/6.x/examples/IPython%20Kernel/Index.ipynb",
          "img": "/static/img/example-nb/ipython-thumb.png"
        },
        {
          "text": "IRuby",
-         "target": "github/SciRuby/sciruby-notebooks/blob/master/getting_started.ipynb",
+         "target": "/github/SciRuby/sciruby-notebooks/blob/master/getting_started.ipynb",
          "img": "/static/img/example-nb/iruby-nb.png"
        },
        {
          "text": "IJulia",
-         "target": "url/jdj.mit.edu/~stevenj/IJulia%20Preview.ipynb",
+         "target": "/url/jdj.mit.edu/~stevenj/IJulia%20Preview.ipynb",
          "img": "/static/img/example-nb/ijulia-preview.png"
        }
      ]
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: nbviewer/frontpage.json [hunk 1]
```diff
@@ -8,17 +8,17 @@
      "links":[
        {
          "text": "IPython",
-         "target": "github/ipython/ipython/blob/6.x/examples/IPython%20Kernel/Index.ipynb",
+         "target": "/github/ipython/ipython/blob/6.x/examples/IPython%20Kernel/Index.ipynb",
```
parente: Have you confirmed the absolute path works when nbviewer is not running as a JupyterHub service?

nbviewer/templates/faq.md
```diff
@@ -4,7 +4,7 @@
 
 <div class="col-md-10 col-md-offset-1">
 
-{% filter markdown(extensions=['headerid(level=3)','toc'], extension_configs= {'toc' : [('anchorlink', True)]}) %}
+{% filter markdown(extensions=['toc'], extension_configs= {'toc' : [('anchorlink', True)]}) %}
```
parente: We should probably specify `markdown>=X` in the `requirements.txt` where X is some minimum version known to work after this code change. The current version you have installed for local dev and test should work. Could you please add it to `requirements.txt` in this PR?

nbviewer/frontpage.json
krinsman: Yes, you're right, it does work also when nbviewer is not running as a JupyterHub service. Sorry for forgetting to mention that.

nbviewer/templates/faq.md
krinsman: That's a good idea. I added it to `requirements.txt`.

parente:
LGTM. Thanks again.

Issue #834, opened 2019-06-09
Update the FAQ to point people to the Discourse


choldgraf:
I just noticed that the FAQ is still pointing people to the `jupyter/help` repository, or the mailing list. We should update those links as `jupyter/help` is (I think) no longer maintained. Discourse seems like a good option.

krinsman:
See https://github.com/jupyter/nbviewer/pull/838

PR #838, opened 2019-06-25
Update the FAQ to point people to the Discourse

krinsman:
This is an implementation of the feature request found in this issue:
https://github.com/jupyter/nbviewer/issues/834


choldgraf:
Nice! Thanks very much

Issue #839, opened 2019-06-25
`nbviewer_config.py` files (similar to JupyterHub)?


krinsman:
At the [Jupyter HPC workshop](https://blog.jupyter.org/jupyter-community-workshop-jupyter-for-scientific-user-facilities-and-high-performance-computing-3afa4a990086) we discussed the idea of implementing a "clone notebook" button to work with JupyterLab, to enable curated notebooks for users to get their feet wet.

@parente This would be very similar to the [nbexamples](https://github.com/danielballan/nbexamples) notebook extension you worked on with Daniel Allan. It seems like good timing to switch to JupyterLab since the [official 1.0 will be released soon](https://discourse.jupyter.org/t/jupyterlab-1-0-release-candidate-published/1438). The current working implementation of this extends NBViewer to have optional buttons to clone the currently viewed notebook into the user's JupyterLab instance, when NBViewer is being run as a JupyterHub service (whence the discussion at a Jupyter HPC workshop).

Based on [preliminary work of Daniel Allan](https://github.com/danielballan/notebook-clone-extension), I have already created a working [notebook cloning extension for NBViewer](https://github.com/krinsman/clonenotebooks). The only wrinkle is that in order to accomplish this, it was necessary to refactor the NBViewer source to make NBViewer extensible.

A PR containing all of the changes to NBViewer ([see the master branch of my fork](https://github.com/krinsman/nbviewer)) needed to make it more extensible would be very large and difficult to review. So I am trying to figure out how to best split it up into a sequence of smaller PR's, each of which will be straightforward to review.

Would a good first step PR be to just include the minimum amount of changes required to make the NBViewer application configurable with an `nbviewer_config.py` file?

krinsman:
See https://github.com/jupyter/nbviewer/pull/840

krinsman:
#840 was merged so I will close this -- sorry for forgetting to do so sooner.

Issue #831, opened 2019-05-10
Request for latest release of nbviewer


krishnatg:
The last released version was back in Oct 2017, and I request the latest code to be released. There is a piece of code from the `providers/local/handlers.py` that deals with following symbolic links which was put in after the last release, which I would like to be part of the nbviewer when I execute a `pip install nbviewer`.

Thanks!

parente:
I've been wanting to cut a release, but felt that we might want to address https://github.com/jupyter/nbviewer/issues/822 first so that we don't have a legacy pinned dependency.

krinsman:
Copy-pasted from my comment in #822 :

> I've fixed this problem in my fork of NBViewer, see here: https://github.com/krinsman/nbviewer/tree/step6
>
> passes tests: https://travis-ci.com/krinsman/nbviewer/builds/120096069
>
> Right now I have a PR open for branch step2, and then this is step6 -- so should I open PR's also for step3, step4, step5, and step6? (Each of which would depend on the previous PR?) Or Would it be better to keep one PR open at a time?

Issue #822, opened 2019-04-07
Fix compatibility with tornado 6


parente:
`gen.Task` no longer exists. I tried quickly porting the logic wholesale into nbviewer but notebook rendering then hung with the notebook fetch promise never resolving. I'm pinning to `tornado<6` in #821 so I can deploy a fix for the jupyter-widgets styling bug. We'll need to revisit tornado compat.

Seanspt:
Same request here. When trying to add nbviewer to our jupyterhub, the following error occurred.
```bash
[E 190513 15:18:16 ioloop:763] Exception in callback functools.partial(<bound method IOLoop._discard_future_result of <tornado.platform.asyncio.AsyncIOMainLoop object at 0x7f58dc07c518>>, <Future finished exception=AttributeError("module 'tornado.gen' has no attribute 'Task'")>)
    Traceback (most recent call last):
      File "/opt/conda/envs/jupyterhub/lib/python3.7/site-packages/tornado/ioloop.py", line 743, in _run_callback
        ret = callback()
      File "/opt/conda/envs/jupyterhub/lib/python3.7/site-packages/tornado/ioloop.py", line 767, in _discard_future_result
        future.result()
      File "/opt/conda/envs/jupyterhub/lib/python3.7/site-packages/tornado/gen.py", line 742, in run
        yielded = self.gen.send(value)
      File "/opt/conda/envs/jupyterhub/lib/python3.7/site-packages/nbviewer/providers/url/client.py", line 80, in _fetch_impl
        response = yield gen.Task(super(NBViewerAsyncHTTPClient, self).fetch_impl, request)
    AttributeError: module 'tornado.gen' has no attribute 'Task'
```

krinsman:
Tornado 6 is incompatible with Python 2.7. So NBViewer needs to drop Python 2.7 compatibility in order to upgrade to Tornado 6. I might be able to work on this if you want, since I have some experience updating to Tornado 6 (which uses Python 3's built-in asyncio module, the first version of Tornado to do so). At least see here: https://github.com/NERSC/sshspawner/pull/14

krinsman:
I've fixed this problem in my fork of NBViewer, see here: https://github.com/krinsman/nbviewer/tree/step6

passes tests: https://travis-ci.com/krinsman/nbviewer/builds/120096069

[...]

PR #840, opened 2019-06-27
Make NBViewer configurable class with 'nbviewer_config.py' file

krinsman:
This PR makes NBViewer a configurable class object, with an 'nbviewer_config.py' file, similar to [JupyterHub](https://github.com/jupyterhub/jupyterhub/blob/master/jupyterhub/app.py).

In terms of code changes, all this does is basically take the `make_app` function and make it the `init_tornado_application` method of the new `NBViewer` class. The `NBViewer` class has a `config_file` attribute added, and then upon constructing the `NBViewer` object config is loaded from `nbviewer_config.py` via the `config_file` attribute, and then `make_app` is run.

See https://github.com/jupyter/nbviewer/issues/839 -- this implements that.

Copy-pasted from there:

> At the [Jupyter HPC workshop](https://blog.jupyter.org/jupyter-community-workshop-jupyter-for-scientific-user-facilities-and-high-performance-computing-3afa4a990086) we discussed the idea of implementing a "clone notebook" button to work with JupyterLab, to enable curated notebooks for users to get their feet wet.

> @parente This would be very similar to the [nbexamples](https://github.com/danielballan/nbexamples) notebook extension you worked on with Daniel Allan. It seems like good timing to switch to JupyterLab since the [official 1.0 will be released soon](https://discourse.jupyter.org/t/jupyterlab-1-0-release-candidate-published/1438). The current working implementation of this extends NBViewer to have optional buttons to clone the currently viewed notebook into the user's JupyterLab instance, when NBViewer is being run as a JupyterHub service (whence the discussion at a Jupyter HPC workshop).

> Based on [preliminary work of Daniel Allan](https://github.com/danielballan/notebook-clone-extension), I have already created a working [notebook cloning extension for NBViewer](https://github.com/krinsman/clonenotebooks). The only wrinkle is that in order to accomplish this, it was necessary to refactor the NBViewer source to make NBViewer extensible.

> A PR containing all of the changes to NBViewer ([see the ~~master~~ `latest` branch of my fork](https://github.com/krinsman/nbviewer)) needed to make it more extensible would be very large and difficult to review. So I am trying to figure out how to best split it up into a sequence of smaller PR's, each of which will be straightforward to review.

> Would a good first step PR be to just include the minimum amount of changes required to make the NBViewer application configurable with an `nbviewer_config.py` file?


krinsman:
The Travis CI tests are incompatible with this PR, because they assume the existence of the `make_app` function in the `app.py` file:

    AttributeError: module 'nbviewer.app' has no attribute 'make_app'

However in this PR the `make_app` function is refactored to become the `init_tornado_application` method of the new `NBViewer` class. This is what [JupyterHub](https://github.com/jupyterhub/jupyterhub/blob/master/jupyterhub/app.py#L2072) does. 

I do like the JupyterHub approach, but even if one didn't, it would still make practical sense to have the same approach between different Jupyter projects, so that contributors to one project would be able to contribute to other projects more easily.

Also the Travis CI tests check for compatibility with Python 2.7, but that contradicts [Jupyter's pledge to sunset Python 2.7](https://python3statement.org/). Moreover it forces NBViewer to [use an outdated version of Tornado](https://www.tornadoweb.org/en/stable/), which is both a security risk and makes Python's built-in asyncio features (which are used extensively in JupyterHub) unavailable. It is also the only reason the Python 2.7 check failed on Travis CI, since this PR uses modern Python 3 syntax (again like other Jupyter projects e.g. JupyterHub).

krinsman:
The reasons for changing the Travis tests to not use `make_app` and Python 2.7 were explained above. 

The IJulia notebook isn't working right now even on `nbviewer.jupyter.org`: [here](https://nbviewer.jupyter.org/url/jdj.mit.edu/~stevenj/IJulia%20Preview.ipynb). The [notebook itself](http://jdj.mit.edu/~stevenj/IJulia%20Preview.ipynb) isn't stored at an HTTPS location; instead the site is HTTP. Right now even trying to access the URL via a browser is impossible (for me). This has happened several times in working on my fork -- the IJulia notebook would just be inaccessible, for no apparent reason. So it doesn't seem to be a very stable website to test on (or to have in the default NBViewer gallery). Anyway I've tested the code with notebooks at numerous other URLs, including the XKCD matplotlib notebook in the default NBViewer gallery, as well as the Nobel-prize-winning LIGO notebook I added to the `test_url.py`. 

In other words I have strong reason to believe, both based on the Travis CI output as well as testing firsthand with `nbviewer.jupyter.org` and my local NBViewer installation that the problem is with the URL and not with this PR. Even going to [jdj.mit.edu](http://jdj.mit.edu/) is impossible.

krinsman:
As far as I can tell, the notebook was created by [Professor Steven G. Johnson at MIT](https://math.mit.edu/~stevenj/), who describes himself as a close collaborator with [Professor John D. Joannopoulos at MIT](http://web.mit.edu/physics/people/faculty/joannopoulos_john.html), to whom the domain `jdj.mit.edu` seems to correspond. The domain [`ab-initio.mit.edu`](http://ab-initio.mit.edu/) also seems to correspond to John D. Joannopoulos, is also served via HTTP instead of HTTPS, and also appears to presently be inaccessible. Perhaps it would make sense to contact Professor Steven Johnson (whose website is served via HTTPS, not HTTP) to ask if he could host his IJulia notebook at his website, instead of Professor John D. Joannopoulos's website, and then link to that new URL in the NBViewer gallery and use that URL instead of the current unreliable `jdj.mit.edu` URL for `test_url.py`.

krinsman:
**Update:** I edited the tests to reflect the issues mentioned above. Professor Joannopoulos's website seems to be working again so I switched the URL test back to the old one. The PR still passes.

parente:
@krinsman Thanks for the recent PRs and issue triage. I hope to have time this weekend to review these.

nbviewer/app.py
```diff
@@ -80,196 +81,200 @@ def nrfoot():
...
+class NBViewer(Application):
 
-def make_app():
-    # NBConvert config
-    config = Config()
-    config.NbconvertApp.fileext = 'html'
-    config.CSSHTMLHeaderTransformer.enabled = False
-    # don't strip the files prefix - we use it for redirects
-    # config.Exporter.filters = {'strip_files_prefix': lambda s: s}
+    config_file = Unicode('nbviewer_config.py', help="The config file to load").tag(config=True)
```
parente: Could you please add a short note in the README about where `nbviewer_config.py` must reside and it's format, maybe via an example of what one might want to customize using the config file?

I don't think this refactor makes NBViewer a first class JupyterApplication supporting the common config outlined here https://jupyter.readthedocs.io/en/latest/projects/config.html, but please correct me if I'm wrong.

.travis.yml
```diff
@@ -4,7 +4,6 @@ node_js:
 - 6
 
 python:
-- 2.7
```
parente: 💯

Could you please update the classifiers in `setup.py` to reflect that the next release will not be 2.7 compatible?

nbviewer/app.py
krinsman: It depends on what one means by a first class JupyterApplication. You are correct that right now the only configurable attribute is the name of the config file. Right now it works the same way JupyterHub's config does -- as long as the file is in the directory from which the nbviewer program is executed, it will pick up the config file. 

I wanted to keep the changes for now as small as possible so that: (1) the chances something which is breaking goes unnoticed are minimized, and (2) the PR is as easy to review as possible. I.e. you are correct, by itself it does mostly nothing, but that's the point. It doing mostly nothing means it's easier to scrutinize, but at the same time still manages to step towards the direction of a first class JupyterApplication.

In order to make NBViewer meaningfully configurable it would be necessary, among other things, to refactor the class to have attributes which are traitlet configurables. In [the next branch of my fork, step2](https://github.com/krinsman/nbviewer/blob/step2/nbviewer/app.py) (a modified version of which I would make as a new PR if this PR was accepted), the `make_app` is refactored so that most of the computed quantities are now attributes (properties) of the NBViewer class. A further PR would then have to identify which of those attributes/properties to make instances of traitlets classes with the `.tag(config=True)` tag (in order to make them configurable from the config file). 

Also another way in which this does not yet make NBViewer a first class JupyterApplication is that it does not yet allow the configurables to be changed by adding command line flags (maybe this is in particular what you had in mind?), so e.g. `python -m nbviewer --c.NBViewer.config_file='better_name.py'` wouldn't work. In JupyterHub, that happens [here](https://github.com/jupyterhub/jupyterhub/blob/3089d441b4d499c2b5662acf8dd8c95837f3490f/jupyterhub/app.py#L183), and is mostly implemented in the traitlets source code [here](https://github.com/ipython/traitlets/blob/master/traitlets/config/application.py#L666). So I do have an idea of how to refactor NBViewer further such that configurable options can be parsed via the command line in that manner, but that for one thing would involve mucking with the 30+ lines of code of `define` function calls currently in `app.py`, which would make the PR hard to read, and which would increase the possibility of accidentally introducing breaking changes.

Regarding what one might want to customize, again with this little PR there really isn't anything to customize yet, you are correct, since there aren't any configurable attributes (besides `config_file` itself). But in later steps (e.g. see [the `latest` branch of my fork](https://github.com/krinsman/nbviewer/blob/latest/nbviewer/app.py#L90), which if this PR is step1, would be around step6 or step7, I think) it's possible to make the handlers for all of the different providers customizable. So, for example, if one wants to subclass the existing handlers, customize them to add extra logic (e.g. for cloning notebooks into a home directory), and then have NBViewer use those custom handlers instead, one can do that. I have the code required for that implementation [here](https://github.com/krinsman/clonenotebooks) (as well as in the `latest` branch of my NBViewer fork); the setup is complicated so I created a [complete Docker image](https://github.com/krinsman/jupyterhub-deploy/tree/master/nbviewer/cloneviewer) which should make it possible to demo locally the type of future that this PR is a step towards, just by running `build.sh`.

Sorry, I've been working on this for about 1.5 months now, it's hard for me to talk about it succinctly since I've thought about it way too much.

.travis.yml
krinsman: I will try to do this as well as the other request (update the README) either this evening or ASAP tomorrow morning. Sorry for the slow reply on my part; I got distracted by my other project besides this at work.

nbviewer/app.py
parente: What you added to the README are the details I was hoping to surface in this PR. Thanks!

parente:
LGTM now. Let's merge it. I'm not planning to redeploy the public instance with these changes just yet, but we'll need to do that at some point just to make sure there aren't any surprises.

krinsman:
@parente If ever/whenever you have time to read this, re: your earlier concern about making NBViewer a first-class Jupyter application, that's accomplished in the latest branch of the fork: https://github.com/krinsman/nbviewer/tree/step7 ([tests passing](https://travis-ci.com/krinsman/nbviewer/builds/120548146))
That's one step ahead of [step6](https://github.com/krinsman/nbviewer/tree/step6), the one which fixes the Tornado compatibility issue mentioned in #831 and #822 .

PR #842, opened 2019-07-13
Reorganize init_tornado_application, create some configurables

krinsman:
Reorganized `init_tornado_application` by creating class properties for the `NBViewer` class, as well as a couple of attributes which are configurable traitlets (and thus can be modified from `nbviewer_config.py`). 

The main principle was: if the variable which was later used in `settings` depended on one of the `options.foo`, then for now the correct way to configure it should be exclusively through the command line (since the command line configuration is not yet compatible with traitlets), and so it should be made a property of the `NBViewer` class. 

Otherwise, it should be made a configurable traitlets instance, with a default value computed using the `@default` decorator from traitlets. 

However, I was not able to apply this principle to to `client` (using an instance of `Any`), since when I did so NBViewer was unable to start (it complained of being unable to find `settings['hub_base_url']` when trying to compute the value of the `hub_base_url` property of `BaseHandler`, even though `hub_base_url` _was_ an entry of `settings`, which suggested that somehow something was fundamentally wrong and the settings were not being correctly understood by the handlers). So in violation of the general rule above I made it a computed property instead.

Also I alphabetized some of the lines of code using Vim sort so that it might be quicker to find things, although admittedly it probably makes the diff more difficult to read.

Anyway, at least in principle, none of the code is changed, and instead it has just been reorganized, exploiting the existence of the new `NBViewer` class.

### DIFF
--- a/nbviewer/app.py
+++ b/nbviewer/app.py
@@ -44,6 +44,11 @@ from .ratelimit import RateLimiter
 from .log import log_request
 from .utils import git_info, jupyter_info, url_path_join
 
+try: # Python 3.8
+    from functools import cached_property
+except ImportError:
+    from .utils import cached_property
+
 #-----------------------------------------------------------------------------
 # Code
 #-----------------------------------------------------------------------------
@@ -115,13 +120,13 @@ class NBViewer(Application):
     def _load_static_url_prefix(self):
         return url_path_join(self.base_url, '/static/')
 
-    @property
+    @cached_property
     def base_url(self):
         # prefer the JupyterHub defined service prefix over the CLI
         base_url = os.getenv("JUPYTERHUB_SERVICE_PREFIX", options.base_url)
         return base_url
 
-    @property
+    @cached_property
     def cache(self):
         memcache_urls = os.environ.get('MEMCACHIER_SERVERS', os.environ.get('MEMCACHE_SERVERS'))
         # Handle linked Docker containers
@@ -154,14 +159,14 @@ class NBViewer(Application):
 
     # for some reason this needs to be a computed property,
     # and not a traitlets Any(), otherwise nbviewer won't run
-    @property
+    @cached_property
     def client(self):
         AsyncHTTPClient.configure(HTTPClientClass)
         client = AsyncHTTPClient()
         client.cache = self.cache
         return client
 
-    @property
+    @cached_property
     def env(self):
         env = Environment(loader=FileSystemLoader(self.template_paths), autoescape=True)
         env.filters['markdown'] = markdown.markdown
@@ -180,7 +185,7 @@ class NBViewer(Application):
 
         return env
 
-    @property
+    @cached_property
     def fetch_kwargs(self):
         fetch_kwargs = dict(connect_timeout=10,)
         if options.proxy_host:
@@ -194,13 +199,13 @@ class NBViewer(Application):
 
         return fetch_kwargs
 
-    @property
+    @cached_property
     def formats(self):
         formats = configure_formats(options, self.config, log.app_log)
         return formats
 
     # load frontpage sections
-    @property
+    @cached_property
     def frontpage_setup(self):
         with io.open(options.frontpage, 'r') as f:
             frontpage_setup = json.load(f)
@@ -213,7 +218,7 @@ class NBViewer(Application):
                               }
         return frontpage_setup
 
-    @property
+    @cached_property
     def pool(self):
         if options.processes:
             pool = ProcessPoolExecutor(options.processes)
@@ -221,12 +226,12 @@ class NBViewer(Application):
             pool = ThreadPoolExecutor(options.threads)
         return pool
 
-    @property
+    @cached_property
     def rate_limiter(self):
         rate_limiter = RateLimiter(limit=options.rate_limit, interval=options.rate_limit_interval, cache=self.cache)
         return rate_limiter
 
-    @property
+    @cached_property
     def template_paths(self):
         template_paths = pjoin(here, 'templates')
         if options.template_path is not None:
--- a/nbviewer/app.py
+++ b/nbviewer/app.py
@@ -25,7 +25,7 @@ from tornado.options import define, options
 
 from jinja2 import Environment, FileSystemLoader
 
-from traitlets import Unicode
+from traitlets import Unicode, Any, Set, default
 from traitlets.config import Application
 
 from .handlers import init_handlers
@@ -74,83 +74,96 @@ FRONTPAGE_JSON = os.path.join(this_dir, "frontpage.json")
 
 class NBViewer(Application):
 
+    name = Unicode('nbviewer')
+
     config_file = Unicode('nbviewer_config.py', help="The config file to load").tag(config=True)
 
-    def init_tornado_application(self):
-        # NBConvert config
-        self.config.NbconvertApp.fileext = 'html'
-        self.config.CSSHTMLHeaderTransformer.enabled = False
+    url_handler         = Unicode(default_value="nbviewer.providers.url.handlers.URLHandler",           help="The Tornado handler to use for viewing notebooks accessed via URL").tag(config=True)
+    local_handler       = Unicode(default_value="nbviewer.providers.local.handlers.LocalFileHandler",   help="The Tornado handler to use for viewing notebooks found on a local filesystem").tag(config=True)
+    github_blob_handler = Unicode(default_value="nbviewer.providers.github.handlers.GitHubBlobHandler", help="The Tornado handler to use for viewing notebooks stored as blobs on GitHub").tag(config=True)
+    github_tree_handler = Unicode(default_value="nbviewer.providers.github.handlers.GitHubTreeHandler", help="The Tornado handler to use for viewing directory trees on GitHub").tag(config=True)
+    gist_handler        = Unicode(default_value="nbviewer.providers.gist.handlers.GistHandler",         help="The Tornado handler to use for viewing notebooks stored as GitHub Gists").tag(config=True)
+    user_gists_handler  = Unicode(default_value="nbviewer.providers.gist.handlers.UserGistsHandler",    help="The Tornado handler to use for viewing directory containing all of a user's Gists").tag(config=True)
 
-        # DEBUG env implies both autoreload and log-level
-        if os.environ.get("DEBUG"):
-            options.debug = True
-            logging.getLogger().setLevel(logging.DEBUG)
-    
-        # setup memcache
-        mc_pool = ThreadPoolExecutor(options.mc_threads)
-    
-        # setup formats
-        formats = configure_formats(options, self.config, log.app_log)
-    
-        if options.processes:
-            pool = ProcessPoolExecutor(options.processes)
-        else:
-            pool = ThreadPoolExecutor(options.threads)
-    
-        memcache_urls = os.environ.get('MEMCACHIER_SERVERS',
-            os.environ.get('MEMCACHE_SERVERS')
-        )
-    
-        # Handle linked Docker containers
-        if(os.environ.get('NBCACHE_PORT')):
-            tcp_memcache = os.environ.get('NBCACHE_PORT')
-            memcache_urls = tcp_memcache.split('tcp://')[1]
-    
-        if(os.environ.get('NBINDEX_PORT')):
+    index = Any().tag(config=True)
+    @default('index')
+    def _load_index(self):
+        if os.environ.get('NBINDEX_PORT'):
             log.app_log.info("Indexing notebooks")
             tcp_index = os.environ.get('NBINDEX_PORT')
             index_url = tcp_index.split('tcp://')[1]
             index_host, index_port = index_url.split(":")
-            indexer = ElasticSearch(index_host, index_port)
         else:
             log.app_log.info("Not indexing notebooks")
             indexer = NoSearch()
-    
+        return indexer
+
+    # cache frontpage links for the maximum allowed time
+    max_cache_uris = Set().tag(config=True)
+    @default('max_cache_uris')
+    def _load_max_cache_uris(self):
+        max_cache_uris = {''}
+        for section in self.frontpage_setup['sections']:
+            for link in section['links']:
+                max_cache_uris.add('/' + link['target'])
+        return max_cache_uris
+
+    static_path = Unicode(default_value=pjoin(here, 'static')).tag(config=True)
+
+    static_url_prefix = Unicode().tag(config=True)
+    @default('static_url_prefix')
+    def _load_static_url_prefix(self):
+        return url_path_join(self.base_url, '/static/')
+
+    @property
+    def base_url(self):
+        # prefer the JupyterHub defined service prefix over the CLI
+        base_url = os.getenv("JUPYTERHUB_SERVICE_PREFIX", options.base_url)
+        return base_url
+
+    @property
+    def cache(self):
+        memcache_urls = os.environ.get('MEMCACHIER_SERVERS', os.environ.get('MEMCACHE_SERVERS'))
+        # Handle linked Docker containers
+        if os.environ.get('NBCACHE_PORT'):
+            tcp_memcache = os.environ.get('NBCACHE_PORT')
+            memcache_urls = tcp_memcache.split('tcp://')[1]
         if options.no_cache:
             log.app_log.info("Not using cache")
             cache = MockCache()
         elif pylibmc and memcache_urls:
+            # setup memcache
+            mc_pool = ThreadPoolExecutor(options.mc_threads)
             kwargs = dict(pool=mc_pool)
-            username = os.environ.get('MEMCACHIER_USERNAME', '')
-            password = os.environ.get('MEMCACHIER_PASSWORD', '')
+            username = os.environ.get("MEMCACHIER_USERNAME", "")
+            password = os.environ.get("MEMCACHIER_PASSWORD", "")
             if username and password:
                 kwargs['binary'] = True
                 kwargs['username'] = username
                 kwargs['password'] = password
                 log.app_log.info("Using SASL memcache")
             else:
-                log.app_log.info("Using plain memecache")
-    
-            cache = AsyncMultipartMemcache(memcache_urls.split(','), **kwargs)
+                log.app_log.info("Using plain memcache")
+
+            cache = AsyncMultiPartMemcache(memcache_urls.split(','), **kwargs)
         else:
             log.app_log.info("Using in-memory cache")
             cache = DummyAsyncCache()
-    
-        # setup tornado handlers and settings
-    
-        template_paths = pjoin(here, 'templates')
-    
-        if options.template_path is not None:
-            log.app_log.info("Using custom template path {}".format(
-                options.template_path)
-            )
-            template_paths = [options.template_path, template_paths]
-    
-        static_path = pjoin(here, 'static')
-        env = Environment(
-            loader=FileSystemLoader(template_paths),
-            autoescape=True
-        )
+
+        return cache
+
+    # for some reason this needs to be a computed property,
+    # and not a traitlets Any(), otherwise nbviewer won't run
+    @property
+    def client(self):
+        AsyncHTTPClient.configure(HTTPClientClass)
+        client = AsyncHTTPClient()
+        client.cache = self.cache
+        return client
+
+    @property
+    def env(self):
+        env = Environment(loader=FileSystemLoader(self.template_paths), autoescape=True)
         env.filters['markdown'] = markdown.markdown
         try:
             git_data = git_info(here)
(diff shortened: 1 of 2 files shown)


krinsman:
@parente If it helps to explain any of the changes, @shreddd said it would be possible to schedule a Zoom call.

parente:
@krinsman I've been heads-down at my day job and not attending to open source much. I'll try to have a look in the near future, or perhaps someone else will.

nbviewer/app.py
```diff
@@ -159,106 +172,138 @@ def init_tornado_application(self):
...
+        # additional settings
+        settings.update(
+                  google_analytics_id=os.getenv('GOOGLE_ANALYTICS_ID'),
+                  gzip=True,
+                  hub_api_token=os.getenv('JUPYTERHUB_API_TOKEN'),
+                  hub_api_url=os.getenv('JUPYTERHUB_API_URL'),
+                  hub_base_url=os.getenv('JUPYTERHUB_BASE_URL'),
+                  log_function=log_request,
+        )
+        log.app_log.info(settings['hub_base_url'])
```
parente: If the intention is to keep this log statement, it would be good to prefix it with `JupyterHub Base URL: %s` or some such.

nbviewer/app.py
```diff
@@ -159,106 +172,138 @@ def init_tornado_application(self):
...
+                  localfile_any_user=options.localfile_any_user,
+                  localfile_follow_symlinks=options.localfile_follow_symlinks,
+                  localfile_path=os.path.abspath(options.localfiles),
+                  mathjax_url=options.mathjax_url,
+                  provider_rewrites=options.provider_rewrites,
+                  providers=options.providers,
+                  render_timeout=options.render_timeout,
+                  statsd_host=options.statsd_host,
+                  statsd_port=options.statsd_port,
+                  statsd_prefix=options.statsd_prefix,
```
parente: Is the goal ultimately to stop using tornado `options` for command line parsing and switch completely to using traitlet configuration on the command line (e.g., `--NBViewer.cache_expiry_max=12345`) and/or in a config file?

I'm asking to understand:

1. if nbviewer will always have a dependence on two configuration systems going forward: tornado.options and traitlets.
2. if the goal is to move everything to traitlets, then when should we cut a final 1.x pypi release of nbviewer fixes up to that point, because we're about to drastically change the command line "API".

I'm sure that seeing the rest of the PRs you have queued up will help answer the question but I figured I'd get these thoughts out as part of the review.

nbviewer/app.py
```diff
@@ -159,106 +172,138 @@ def init_tornado_application(self):
...
+    def template_paths(self):
+        template_paths = pjoin(here, 'templates')
+        if options.template_path is not None:
+            log.app_log.info("Using custom template path {}".format(options.template_path))
+            template_paths = [options.template_path, template_paths]
+
+        return template_paths
+
+
+    def init_tornado_application(self):
```
parente: I agree with your comment that it's hard to determine if all of the old command line parameters are having the same net effect after the refactor by reading the diff alone. I wonder if it would be possible to add a test that vets what's in the `settings` dict is what is expected (e.g., before and after these changes).

nbviewer/app.py
krinsman: Oh oops no that was a goof. Had that there when I was debugging something but forgot to remove it. Sorry about that.

nbviewer/app.py
krinsman: The goal is to move everything to traitlets (see https://github.com/jupyter/nbviewer/pull/847 where this was finally done). 

I was actually able (I think) to move everything to traitlets without affecting the command line "API", so hopefully that won't be an issue. (It's possible using aliases and flags.)

This would make NBViewer a first-class Jupyter application like you mentioned, plus the configuration system is more extensible and flexible, since now one can use both the command line as well as the config file.

krinsman: So e.g. you can do `--cache_expiry_max=12345` _or_ `--NBViewer.cache_expiry_max=12345`. (As well as add a line `c.NBViewer.cache_expiry_max = 12345` to `nbviewer_config.py`, with the command line settings taking precedence over the config file ones. It's even possible to change the config file via the command line.)

I'm not sure what a "cut a release" means or what that would entail. Sorry for being dumb; I haven't ever been this involved in an OSS project before and am probably not familiar with all of the terminology.

nbviewer/app.py
krinsman: Yes that sounds like it would be a good idea. 

I don't think I know how to do that. I don't really understand or know much about `unittests.mock` or `nosetests`. Any suggestions for what to write would be appreciated.

At the very least _everything_ has worked all of the (dozens maybe even hundreds) times I tested things "interactively" (for lack of a better word, i.e. running the program in the Docker container, as a standalone application and as a JupyterHub service, viewing all of the example notebooks, viewing GitHub users, repos, user gists, slides, etc). But I know this isn't enough to test the more advanced features like `threads` or `pools` or `memcache`. I'm not sure if I even have all of the hardware and/or software necessary to test those. I tried to ensure that I was just moving code verbatim as much as possible if I didn't fully understand it, and not inadvertently changing it. But the git diff is largely useless for verifying that, since it just registers things which are cut and pasted to another portion of the file as deleted.

nbviewer/app.py
krinsman: Deleted that -- thank you for pointing this out!

nbviewer/app.py
parente: I forgot about the alias support in traitlets. I believe you're correct that we can switch to traitlets without intentionally changing the CLI parameters. There might be some gotchas in how things are parsed, but we can deal with those as we find them.

> I'm not sure what "cut a release" means or what that would entail. 

By "cut a release" I meant making a tag here on GitHub and pushing a package to PyPI before we make any substation changes to the CLI. However, given that you're using aliases, that might not be necessary to do up front.

> Sorry for being dumb; I haven't ever been this involved in an OSS project before and am probably not familiar with all of the terminology.

You're not being dumb and you're doing great. Thank you for all the work!

nbviewer/app.py
parente: I'll need to take a look at the test suite to see how easily we can put something in. However, now that you reminded me about aliases, it might not be worth the effort.

parente:
@krinsman I'll try to allocate time each night this week to go through your PRs. Thank you again for submitting them!

nbviewer/app.py
```diff
@@ -74,83 +74,96 @@ def nrfoot():
...
+    local_handler       = Unicode(default_value="nbviewer.providers.local.handlers.LocalFileHandler",   help="The Tornado handler to use for viewing notebooks found on a local filesystem").tag(config=True)
+    github_blob_handler = Unicode(default_value="nbviewer.providers.github.handlers.GitHubBlobHandler", help="The Tornado handler to use for viewing notebooks stored as blobs on GitHub").tag(config=True)
+    github_tree_handler = Unicode(default_value="nbviewer.providers.github.handlers.GitHubTreeHandler", help="The Tornado handler to use for viewing directory trees on GitHub").tag(config=True)
+    gist_handler        = Unicode(default_value="nbviewer.providers.gist.handlers.GistHandler",         help="The Tornado handler to use for viewing notebooks stored as GitHub Gists").tag(config=True)
+    user_gists_handler  = Unicode(default_value="nbviewer.providers.gist.handlers.UserGistsHandler",    help="The Tornado handler to use for viewing directory containing all of a user's Gists").tag(config=True)
```
parente: The `--providers` option on the command line already provides a good amount of control over selecting which providers in any dotted package (e.g., `foo.bar.my.providers`) should be loaded and their `default_handlers` registered. Is there a benefit to calling out these specific handlers as configurable attributes?

nbviewer/app.py
```diff
@@ -159,106 +172,137 @@ def init_tornado_application(self):
...
-                               'show_input': True,
-                               'sections': frontpage_setup}
-    
-        # cache frontpage links for the maximum allowed time
-        max_cache_uris = {''}
-        for section in frontpage_setup['sections']:
-            for link in section['links']:
-                max_cache_uris.add('/' + link['target'])
-    
+        env.globals.update(nrhead=nrhead, nrfoot=nrfoot, git_data=git_data, jupyter_info=jupyter_info(), len=len)
```
parente: I'm wondering about the use of properties throughout the new code. 

In the original code, various values were computed and objects instantiated once-and-only-once in the top-to-bottom flow of `init_tornado_application`. Now many of the values have the potential to be computed more than once if their properties are accessed more than one time. For example, if we forget or a new developer doesn't know that`self.env` is a property, each time it's referenced the code will create a fresh instance of a `jinja2.Environment` and potentially configure it in a different manner. 

Maybe it doesn't matter if we build a new `env` each time. But I'm perhaps it will in the future. Or perhaps it does already matter for other properties (`pool`?) I suspect these properties acting like factory functions will become hard to maintain and understand over time.

Two thoughts on what we could do instead:

1. Have the properties cache their computed values on first access and only return the cached value there after.
2. Make the properties regular functions so that it's clear an explicit action is being taken each time one is called.

What do you think?

nbviewer/app.py
krinsman: My understanding is that the `--providers` option, compared to the defaults, currently only allows the user to choose a subset of the available providers. The idea of this isn't to do that; instead it's to allow the user to choose different programs entirely to be the providers, regardless of whatever subset of providers (GitHub, Local, etc.) they might choose.

For example the thing I made as a followup to `nbexamples` ([see here](https://github.com/krinsman/clonenotebooks/blob/master/clonenotebooks/renderers/renderers.py)) works in part by sub-classing the handlers in the NBViewer code and rewriting some of their methods. Then in the config the user has to specify using these custom handlers instead of the handlers in the NBViewer source (e.g. use a custom handler to view GitHub).

I think in this PR right now these lines are a no-op since there isn't actually any way for NBViewer to know what to do with this information. 

But in the nextPR (or the one after that), the code is rewritten to use the string e.g. 
```nbviewer.providers.github.handlers.GitHubBlobHandler```
 to do (dynamically) the equivalent of 
```from nbviewer.providers.github.handlers import GitHubBlobHandler```
with the idea being that one could instead also dynamically import
```from clonenotebooks.renderers import GitHubBlobRenderingHandler```
by setting the value of `github_blob_handler` to `clonenotebooks.renderers.GitHubBlobRenderingHandler`.

In short I think this will implement new functionality, but maybe I misunderstood the code.

File: nbviewer/app.py (hunk 5 above)
Comment (krinsman) [reply nbviewer/app.py]: My understanding was that properties are only computed once, so that they don't implement an observer pattern by default. So I thought this was logically equivalent to the original code. At the very least (I think) I know that for traitlets, the very similar `@default` decorator only computes the value once, and that one has to manually add an `@observe` decorator to implement an observer pattern. So maybe I should change these to traitlets?

I feel like I'm misunderstanding the concern.

Comment (krinsman) [reply nbviewer/app.py]: I guess I should have used setters instead of properties? https://www.python-course.eu/python3_properties.php

My rough understanding was that a property could safely be thought of as an attribute whose (initial and final for all time) value could be computed dynamically, using the function decorated by the decorator, and that therefore the property was effectively "read-only". But the more I look into this, the less this seems like this is the case. 

I know that traitlets attributes can be over-written outside of the class where they are defined, but they are at least "read-only" within the class where they are defined. (At least that has been my practical experience.) So maybe I should make these traitlets?

Comment (krinsman) [reply nbviewer/app.py]: Oh crud yeah I didn't understand how properties work at all -- the function is actually called every time the property is accessed, not just once: https://wiki.python.org/moin/ComputedAttributesUsingPropertyObjects

This behavior wasn't intentional on my part. I think option **1.** you listed above is the better one (if for no other reason than that was what I had originally intended anyway). Can that be implemented using traitlets and the `@default` decorator? 

I'm not sure how else that might be done. I thought the purpose of properties was for this sort of code organization, not to recompute the value every single time.

Comment (krinsman) [reply nbviewer/app.py]: Also if properties work that way, then it would seem like they automatically implement the observer pattern. But I thought I read elsewhere that properties on their own weren't sufficient to implement the observer pattern, which I think was part of what led to my mistaken belief about how they work. So basically now I'm confused. I know that for traitlets one has to also use the `@observe` decorator, not just `@default`, to implement the observer pattern, but that's about it.

Comment (krinsman) [reply nbviewer/app.py]: Should I use this? https://docs.python.org/dev/library/functools.html#functools.cached_property
Apparently it requires Python 3.8 though which in practice probably isn't a realistic dependency.

File: nbviewer/app.py (hunk 4 above)
Comment (parente) [reply nbviewer/app.py]: I believe it's possible already to pass `--providers my.package.whatever` and include a `default_handlers = [MyOwnGitBlobHandler]` in that `whatever.py` to have nbviewer load a custom handler of your choosing. I agree it's a bit awkward to say "load my handler and all the others except for `nbviewer.providers.github.handlers.GitHubBlobHandler` without specifying everything in a whitelist, but unless I'm misunderstanding the code today, I think the ability to add new handlers from arbitrary packages already exists and is more extensible than naming the N default handlers in their own traitlets.

I'll try to craft an example to prove out this hypothesis this evening.

Comment (krinsman) [reply nbviewer/app.py]: Yes I agree, having to have a variable for every single handler isn't a good approach. I hope there is a better one already built-in.

File: nbviewer/app.py (hunk 5 above)
Comment (krinsman) [reply nbviewer/app.py]: OK I updated the commit to change the properties to (what should be according to StackOverflow) cached properties.

File: nbviewer/app.py (hunk 4 above)
Comment (krinsman) [reply nbviewer/app.py]: My understanding was that the values for the fields of `--providers` were lists of strings, and the only possible strings were `nbviewer.providers.github`, `nbviewer.providers.gist`, `nbviewer.providers.url`, `nbviewer.providers.local`. In particular, my understanding was that there isn't any way to choose which handler is used for each provider. 

This is an issue in particular for GitHub and Gist, where there isn't actually a one-to-one correspondence between rendering handlers and providers (unlike for URL and local); GitHub has both `GitHubBlobHandler` and `GitHubTreeHandler` (and `GitHubUserHandler`), whereas Gist has both `GistHandler` and `UserGistsHandler`. Thus it's unclear to me how the interface allows the user to change e.g. only `GitHubBlobHandler` and not `GitHubTreeHandler`. 

The naming convention however seems unclear though, since if including the string `'nbviewer.providers.github'` just means "include the GitHub provider", then why not have the string just be `'github'`? It seems like the `nbviewer.providers` prefix provides no useful information, especially since all providers (including any hypothetical future ones) are located there.

Comment (parente) [reply nbviewer/app.py]: I didn't have time last night to dig in. About to do that now.

To be clear, I'm looking to avoid enshrining the GitHub, Gist, local, etc. providers with some special configuration status. I believe the original intent was to treat them as a set of defaults that happen to be included in nbviewer itself, all of which could be disabled / replaced with an entirely different set implemented and shipped outside nbviewer.

Comment (krinsman) [reply nbviewer/app.py]: That sounds like a good idea. I'm somewhat confused though, because being able to disable/replace them with entirely different sets sounds like the same thing as them being configurable, so maybe we are using the same words to mean different things?

Anyway I have no personal preference regarding how they can be disabled/replaced, so if it's already built-in, then that works just as well and is definitely what we should do.

Comment (parente) [reply nbviewer/app.py]: https://github.com/parente/nbviewer-provider demonstrates a (simple/silly) custom URI rewrite rule and associated provider installed in a separate package from nbviewer. The README in the repo has some pointers about to run it and where to look for interesting bits.

I believe the example shows that nbviewer already supports the addition or subtraction of arbitrary handlers without calling out gist, github, url, etc. as special built-in types of handlers to be configured via specific traitlets. 

What I don't know is if the providers / provider rewrites satisfy your use case or not, or if switching to them is a major speed bump in the path of your of your PRs. Let's try to discuss a bit more here, and maybe switch to a video chat one day if we can't easily find a path forward.

Comment (krinsman) [reply nbviewer/app.py]: **TL;DR** I don't understand at all how the `provider_rewrites` or `uri_rewrites` code already in NBViewer works, or what it is supposed to do, or even what the term "URI rewrite" means. I believe you that it may be suitable for my use case, but I don't understand the API behind how it works or etc. So I can't say either way at this point. 

I will need to spend some time digging deeper into how it works and what it does before being able to tell you if it will work or not.

For the record switching to them is almost certainly a major speed bump in the path of my PR's, but I doubt that's a good reason for me to not switch to them. I mean, it's a good reason for me personally, out of sheer laziness, but I don't see how it would be good for the project. (Assuming, again, that it is even possible, which I am not yet sure of either way.)
---
I think it might, I'm not sure at this point yet. I saw the uri_rewrites functions in the code but never really understood what they do. I know that URI is a somewhat more general term than URL but otherwise don't really have enough understanding of web technology to have intuition for the term.

Looking at the example my initial concern is that it can only append routes, but not actually rewrite them. My use case requires (at least I think) completely over-riding the built-in handlers and sending the requests to custom handlers. The example looks like it only supports adding new handlers/providers whose routes don't conflict with the existing ones.

At least my understanding of a Tornado webserver is that if two handlers have conflicting routes, i.e. if a URL matches the regexes in the routes for two distinct URLSpecs (corresponding to two distinct handlers), the URLSpec which was added to the webserver first gets priority, which in this case would be the built-in provider handlers.

The key file seems to be here: https://github.com/jupyter/nbviewer/blob/master/nbviewer/providers/__init__.py
I guess I don't really understand what `provider_uri_rewrites` does because it basically just calls `_load_provider_feature` with `feature='uri_rewrites'`, but that parameter in the `_load_provider_feature` is only used in the call to `__import__`, specifically so that it is `__import__(module, fromlist=[feature])`. At least in my admittedly crude attempts at experimenting with this, it doesn't really seem like this does anything (like if the desired thing is in the `__init__.py` file, it will get imported anyway without having to add `fromlist=[feature]`.

So again, my crude understanding is that if the `fromlist` parameter of `__import__` doesn't do anything, and all `provider_uri_rewrites` does is change the value of that parameter, then seemingly that function doesn't do anything. And even if it does do something, the only place I could find the `provider_uri_rewrites` function used in the code is in one of the methods of `CreateHandler` (https://github.com/jupyter/nbviewer/blob/master/nbviewer/handlers.py#L63) which itself does not seem to do anything. (The comment for that class says "handle creation via frontpage form", but creation of what? And what frontpage form? Would the user have to create the frontpage form themselves? If so, how would they do that?)

Also another comment in that class says "storing this on a class attribute is a little icky", but what is "_this_"?

That being said the comment that "providers is a list of module import paths" does seem to suggest it does what you claim it does. (This also appears to be the place where the `provider_rewrites` setting you modified via the command line in the Dockerfile/Makefile finally gets used, since it isn't used in app.py, just passed to the handlers via the `settings` dict.)

I think part of my confusion stems from the fact that at least in practice I only know Python 3, not Python 2. So some of the code which seems it might be used for compatibility reasons with Python 2 (e.g. the `fromlist` parameter in `__import__`, or the `type(self).uri_rewrite_list = provider_uri_rewrites(providers)`) might be confusing me. Although more likely I am just not very good at Python 3, and all of that is syntactically necessary Python 3 as well.

My personal preference, out of sheer laziness if no other reason, would be to ignore the `provider_rewrites` entirely, since that is probably the one part of the code base which was and remains most mysterious to me. But I think your point is valid that it doesn't make sense to duplicate a feature which already exists. 

(Although the devil on my shoulder says that making all of these traitlets configurables is more analogous to what JupyterHub does with the Spawner and Authenticator classes among others, and according to GitHub search JupyterHub doesn't have any functions of variables with names including `rewrite`, `rewrites`, or `uri_rewrites`, so to the extent it implements any similar functionality it does it in a different way which maybe we should copy instead, so that it's easier for people to work on both NBViewer and JupyterHub. Regarding adding new or extra providers, that seemingly could be accomplished by using an `extra_handlers` Traitlet configurable like JupyterHub has: https://github.com/jupyterhub/jupyterhub/blob/master/jupyterhub/app.py#L1198 which could be specified now entirely via config file.)

Comment (krinsman) [reply nbviewer/app.py]: @parente OK I think I am starting to get now how it's supposed to work.

So if I want to make my own custom GitHub provider/handlers (one of my use cases), do I have to copy-paste the `default_handlers` and the `uri_rewrites` functions into a `handlers.py` file which I then specify currently using the command line argument? Except that I have to manually change all of the lines in `default_handlers` to the name of my custom handler?

Comment (parente) [reply nbviewer/app.py]: > So if I want to make my own custom GitHub provider/handlers (one of my use cases), do I have to copy-paste the `default_handlers` and the `uri_rewrites` functions into a handlers.py file which I then specify currently using the command line argument? 

The module you name on the command line as a value for `--providers` has to implement a `default_handlers` function, a `uri_rewrites` function, or both.

> Except that I have to manually change all of the lines in default_handlers to the name of my custom handler?

The `default_handlers` function is expected to return a growing list of tornado handlers. The function receives a list of handlers registered ahead of this function by other `--provider` modules specified at startup. The function returns another list which, typically but not necessarily, contains one or more additional handlers.

> But what if I only want to use my custom GitHub handler, but keep the default URL, Gist, and local handlers?

I think the following is the correct syntax for the tornado CLI parser, but you've got the general idea:

```
python -m nbviewer --providers='customgithubhandler' \
--providers="nbviewer.providers.url" \
--providers="nbviewer.providers.local" \
--providers="nbviewer.providers.gist" 
--provider_rewrites="customgithubhandler" \
--provider_rewrites="nbviewer.providers.url" \
--provider_rewrites="nbviewer.providers.local" \
--provider_rewrites="nbviewer.providers.gist" 
```

This is the awkwardness I mentioned in an earlier comment, one that I think can be assuaged by something like:

```
python -m nbviewer --built-in-providers \
--providers='customgithubhandler' \
--provider_rewrites="customgithubhandler"
```

if you wanted to register all the built-in providers and then simply extend that original set with your own.

> So the fields in provider_rewrites and providers are strings specifying (sub)modules whose corresponding directories contain an __init__.py file which imports in the former case a function called uri_rewrites and in the latter case a function called default_handlers?

The values are modules or packages located using normal Python module resolution.

> I don't think this is necessarily ideal though in the case where one is just making a custom version of the built-in handlers/providers -- in that case the user will want all of the code to be identical, except for a couple of lines where e.g. GitHubBlobHandler is changed to GitHubBlobRenderingHandler. And so every time NBViewer makes a slight change to that or to that routes supported (e.g. like the PR I made recently with the GitHub URL fix), the custom handler will break.

> The model I have in mind is like in JupyterHub, where there is a default Spawner or Authenticator class defined in the code, but users can subclass those, create their own, and then have the subclassed ones imported dynamically. But the important point being that the subclassed Spawners/Authenticators are used identically by JupyterHub to how the built-in ones are used. And also in this case the idea is to supplement/add on to the built-ins, not create new handlers or etc.

I think this is the crux of the conversation. I believe Spawners and Authenticators are both configuration points and base classes meant to be extended with diverse implementations in JupyterHub. In contrast, I believe URLHandler, LocalFileHandler, GitHubBlobHandler, GitHubTreeeHandler *are* those diverse implementations of the BaseHandler class in nbviewer as currently implemented, while `providers` and `provider_rewrites` are the configuration mechanism for specifying the implementations to load.

> What I am trying to say is that admittedly this PR does something very similar to the already existing features, but not the same. I think they are designed for different problems: the already existing features are better for adding new handlers/new providers, whereas the changes in the PR's are for customizing the built-in handlers.

Looking ahead, I think I see that your later PRs make URLHandler, LocalFileHandler, GitHubBlobHandler, GitHubTreeeHandler, etc. more extensible so that they can serve as base classes for yet more implementations. I think those changes are great and we should merge them, full stop. What I think we need to decide for this PR is whether:

1. To continue support the `providers` and `provider_rewrites` as the configuration mechanism for specifying handler implementations and the URL parsing rules that route to them
2. To add new traitlets / configuration options for overriding the default implementations of git tree handling, git blob handling, gist handling, etc. specifically while continuing to support extensibility through `providers` and `provider_rewrites` as well.
3. To choose some other option that is eluding me at the moment.

I think it would be wise to get a some folks who built and maintained nbviewer to weigh in here on the different options before we proceed. Starting at the top of the contributors list, maybe @minrk or @Carreau, if they have time and interest.

It might also help us come to a conclusion if we briefly lay out what a user would have to do using providers/provider_rewrites today or the proposed configuration system after merging all of the outstanding PRs to:

* Enable only the built in local URL handler?
* Enable a custom github blob handler as well as all the other built in handlers?
* Enable only a custom handler from a third-party package?

I can take a stab at the former tomorrow.

krinsman: @parente My thinking when I wrote this, since at least currently it's the option of least resistance (i.e. maximum laziness on my part), would be option 2. But I agree that at least longer-term this probably suggests a refactoring/rethinking of the provider extensibility.

Thinking about this last night, I guess the main issue is that a Tornado URLSpec is a tuple with (up to) three elements, and hypothetically a user might want to do any of:
- customize any one (or more) of the entries of an existing URLSpec in the NBViewer program without changing any of the others.
- add their own URLSpecs
- customize/change the order in which the URLSpecs are added to the application/server.

So basically to rephrase what we seem to agree on is that the subsequent PR's are more about the first bullet point (e.g. the `handler_settings` allows the user to change the third element), whereas the existing `providers`/`provider_rewrites` options are more about the second/third bullet point.

It would be helpful to users and to us if, as you wisely suggested, we think about how to organize all of these levels of customization in a way that is easier to understand and/or requires fewer command line arguments. That would be something good for me to tackle in an existing PR, and/or at least to add a `docs` folder to the repository with some basic Sphinx templates that explain what's currently in NBViewer. The example providers repository you made above seems like it would make a great [cookiecutter template](https://cookiecutter.readthedocs.io/en/latest/) and/or something to add to some docs.

krinsman: @parente Regarding the bullet points, I believe the answers for the first two bullets are:
- `python -m nbviewer --providers='nbviewer.providers.local'`
-     python -m nbviewer --NBViewer.github_blob_handler=mymodule.CustomGitHubBlobHandler --localfiles

(Since the local file handler isn't enabled by default although the other three are.)

For the third my understanding from your example is that it would be:
-     python -m nbviewer --providers=mymodulewithdefaulthandlersandurirewrites --provider_rewrites=mymodulewithdefaulthandlersandurirewrites

Also a lot of these are kind of ugly, but now that there's a config file it won't be necessary to do it all in command line; the changes can be typed out once in the config file and re-loaded automatically every time without additional typing.

> The module you name on the command line as a value for --providers has to implement a default_handlers function, a uri_rewrites function, or both.

Maybe this would allow us to merge the `--provider_rewrites` alias with the `--providers` one? I.e. if someone loads a new provider module using the `--providers` alias, then if the module contains URI rewrites but no new URLSpecs, only the URI rewrites are loaded, if only new URLSpecs but no URI rewrites then only new handlers, if both URLSpecs and URI rewrites then both.

That only addresses the issue of customizing by adding new things however, and not the issue of only selecting a subset of existing things and/or modifying some subset of existing things. But that being said my understanding of the URI rewrites concept is still shaky.

parente: @krinsman, I had to step away from most open source work for a few days. Thanks for posting your thoughts in the meantime.

I think there's something I'm missing about your use case that the traitlets solve easily that the provider configuration does not. I also think the traitlets don't interfere with the existing `provider/provider_rewrites` capability. So rather than going back and forth until I have an epiphany, let's do the following:

1. Merge this PR.
2. Continue with reviewing, adjusting, merging the remaining PRs
3. Revisit the multiple ways of configuring nbviewer providers through documentation, consolidation, or some combination of the two at the end

Does that sound reasonable to you?

krinsman: @parente It definitely sounds reasonable to me.

I have been thinking more about 3., and am currently looking into whether making the URLSpecs themselves the traitlet configurables and/or some nested routing configuration would be a more useful/concise way to organize all of this. But I have not had any epiphanies yet either.

parente:
OK. This PR has sat open without comment from anyone else for a while now. Following through with:

1. Merging this PR.
2. Continue with reviewing, adjusting, merging the remaining PRs
3. Revisiting the multiple ways of configuring nbviewer providers through documentation, consolidation, or some combination of the two at the end

PR #843, opened 2019-07-29
Step 3

krinsman:
Made choice of (provider) handlers customizable via 'nbviewer_config.py'.

Depends on https://github.com/jupyter/nbviewer/pull/842 ("step 2")

### DIFF
--- a/nbviewer/app.py
+++ b/nbviewer/app.py
@@ -25,7 +25,7 @@ from tornado.options import define, options
 
 from jinja2 import Environment, FileSystemLoader
 
-from traitlets import Unicode, Any, Set, default
+from traitlets import Any, Dict, Set, Unicode, default
 from traitlets.config import Application
 
 from .handlers import init_handlers
@@ -82,6 +82,9 @@ class NBViewer(Application):
     name = Unicode('nbviewer')
 
     config_file = Unicode('nbviewer_config.py', help="The config file to load").tag(config=True)
+
+    # Use this to insert custom configuration of handlers for NBViewer extensions
+    handler_settings = Dict().tag(config=True)
 
     url_handler         = Unicode(default_value="nbviewer.providers.url.handlers.URLHandler",           help="The Tornado handler to use for viewing notebooks accessed via URL").tag(config=True)
     local_handler       = Unicode(default_value="nbviewer.providers.local.handlers.LocalFileHandler",   help="The Tornado handler to use for viewing notebooks found on a local filesystem").tag(config=True)
@@ -243,7 +246,16 @@ class NBViewer(Application):
 
     def init_tornado_application(self):
         # handle handlers
-        handlers = init_handlers(self.formats, options.providers, self.base_url, options.localfiles)
+        handler_names = dict(
+                  url_handler=self.url_handler,
+                  github_blob_handler=self.github_blob_handler,
+                  github_tree_handler=self.github_tree_handler,
+                  local_handler=self.local_handler,
+                  gist_handler=self.gist_handler,
+                  user_gists_handler=self.user_gists_handler,
+        )
+        handler_kwargs = {'handler_names' : handler_names, 'handler_settings' : self.handler_settings}
+        handlers = init_handlers(self.formats, options.providers, self.base_url, options.localfiles, **handler_kwargs)
         
         # NBConvert config
         self.config.NbconvertApp.fileext = 'html'
--- a/nbviewer/handlers.py
+++ b/nbviewer/handlers.py
@@ -75,31 +75,49 @@ class CreateHandler(BaseHandler):
 # Default handler URL mapping
 #-----------------------------------------------------------------------------
 
-def format_handlers(formats, handlers):
-    return [
+def format_handlers(formats, urlspecs, **handler_settings):
+    urlspecs = [
         (prefix + url, handler, {
             "format": format,
             "format_prefix": prefix
         })
         for format in formats
-        for url, handler in handlers
+        # Tornado handler URLSpec of form (route, handler_class, initalize_kwargs)
+        # https://www.tornadoweb.org/en/stable/web.html#tornado.web.URLSpec
+        # kwargs passed to initialize are None by default but can be added
+        # https://www.tornadoweb.org/en/stable/web.html#tornado.web.RequestHandler.initialize
+        for url, handler, initialize_kwargs in urlspecs
         for prefix in [format_prefix + format]
     ]
+    for handler_setting in handler_settings:
+        if handler_settings[handler_setting]:
+            # here we modify the URLSpec dict to have the key-value pairs from
+            # handler_settings in NBViewer.init_tornado_application
+            for urlspec in urlspecs:
+                urlspec[2][handler_setting] = handler_settings[handler_setting]
+    return urlspecs
 
+def init_handlers(formats, providers, base_url, localfiles, **handler_kwargs):
+    # `handler_kwargs` is a dict of dicts: first dict is `handler_names`, which
+    # specifies the handler_classes to load for the providers, the second
+    # is `handler_settings` (see comments in format_handlers)
+    # Only `handler_settings` should get added to the initialize_kwargs in the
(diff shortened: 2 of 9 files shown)


parente:
@krinsman Can you rebase this on master now that step 2 has merged?

parente:
Also, master branch build failed after merging the prior PR. It looks unrelated and due to a dependency change. We'll have to address it before merging this one and beyond.

krinsman:
@parente OK I think I rebased it successfully, at least it looks like it comparing the three and two dot diffs.

krinsman:
OK I remember having that problem before earlier in one of my branches (I think step7, where the Travis worked locally for me after the fix but not when I made the PR against the Jupyter repository) : https://stackoverflow.com/questions/46290556/installing-pycurl-with-fatal-error-gnutls-gnutls-h-no-such-file-or-directory

Let me try editing the `.travis.yml` in accordance with the StackOverflow answer (since I remember it working for me locally).

nbviewer/providers/__init__.py
```diff
@@ -21,7 +21,22 @@ def provider_handlers(providers):
        example, custom URLs which should be intercepted before being
        handed to the basic `url` handler
     """
-    return _load_provider_feature('default_handlers', providers)
+    # `handler_kwargs` is a dict of dicts: first dict is `handler_names`, which
+    # specifies the handler_classes to load for the providers, the second
+    # is `handler_settings` (see comments in `format_handlers` in nbviewer/handlers.py)
```
parente: This comment block would make a good docstring describing the `handler_kwargs` parameter.

nbviewer/providers/__init__.py
```diff
@@ -35,17 +50,58 @@ def provider_uri_rewrites(providers):
...
 
-def _load_provider_feature(feature, providers):
+def _load_provider_feature(feature, providers, **handler_names):
     """Load the named feature from an ordered list of dotted-notation modules
        which each implements the feature.
 
        The feature will be passed a list of feature implementations and must
        return that list, suitably modified.
     """
+    # `handler_names` is the same as the `handler_names` attribute of the NBViewer class
```
parente: I suggest moving this line into the docstring describing the parameter.

nbviewer/providers/__init__.py
```diff
@@ -35,17 +50,58 @@ def provider_uri_rewrites(providers):
...
+
+    # Ex: provider = 'nbviewer.providers.url' 
+    # provider.rsplit(',', 1) = ['nbviewer.providers', 'url']
+    # provider_type = 'url'
+    provider_types = [provider.rsplit('.', 1)[-1] for provider in providers]
+
+    if 'github' in provider_types:
+        provider_types.append('github_blob')
+        provider_types.append('github_tree')
+        provider_types.remove('github')
```
parente: Hmm. This function used to be a handful of lines long and agnostic as to any specific providers it was loading. Is there a better place to put this logic specific to loading github providers?

parente: After reading the code here a couple times, I'm having trouble understanding what the new logic in the function is meant to support. Can you summarize the change?

nbviewer/handlers.py
```diff
@@ -75,31 +75,49 @@ def get_provider_rewrites(self):
...
+                urlspec[2][handler_setting] = handler_settings[handler_setting]
+    return urlspecs
+
+def init_handlers(formats, providers, base_url, localfiles, **handler_kwargs):
+    # `handler_kwargs` is a dict of dicts: first dict is `handler_names`, which
+    # specifies the handler_classes to load for the providers, the second
+    # is `handler_settings` (see comments in format_handlers)
+    # Only `handler_settings` should get added to the initialize_kwargs in the
+    # handler URLSpecs, which is why we pass only it to `format_handlers`
+    # but both it and `handler_names` to `provider_handlers`
```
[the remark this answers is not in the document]
krinsman: This is a good idea. I will try to do that now!

nbviewer/app.py
```diff
@@ -243,7 +246,16 @@ def template_paths(self):
...
+        handler_names = dict(
+                  url_handler=self.url_handler,
+                  github_blob_handler=self.github_blob_handler,
+                  github_tree_handler=self.github_tree_handler,
+                  local_handler=self.local_handler,
+                  gist_handler=self.gist_handler,
+                  user_gists_handler=self.user_gists_handler,
+        )
+        handler_kwargs = {'handler_names' : handler_names, 'handler_settings' : self.handler_settings}
+        handlers = init_handlers(self.formats, options.providers, self.base_url, options.localfiles, **handler_kwargs)
```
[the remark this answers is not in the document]
krinsman: I think you're right. I recall having this argument with myself, and I think the reason I chose this approach was essentially laziness. 

At least, assuming we're referring to the same thing, I didn't want to have to remember arguments which were implicitly in the function signature via the kwargs so I wrote them out explicitly in the function definitions. But then when calling the functions/methods, especially when subclassing them, I didn't want to have to remember all of their function signatures, so I would often just pack things in as a dict.

I agree it decreases readability. I think I basically had the sense though that this approach is probably only provisional and going to have to be refactored one day, so I wanted to retain some flexibility and not hard-code too many assumptions about the function signatures (as well as avoid repeatedly getting annoying errors like `keyword argument 'blah' is not defined for 'blah_funch'`).

If you want me to look into a more long-term solution for right now then I will.

nbviewer/providers/__init__.py
krinsman: Ditto the above!

nbviewer/providers/__init__.py
krinsman: The first thing to note was that this was basically a hacky thing I did to ensure backward compatibility, not because I thought it was a good design per se. (Actually this is probably the portion of the code I wrote which I hate the most, but that's another story.)

Basically my crude understanding was that the `providers` command line argument only allows one to specify things at the level of individual "providers", which to me seemed like a unit of abstraction encompassing an entire "media source" with accompanying submodule containing multiple handlers and URI rewrites. Whereas what I was trying to write was at the sub-level of individual handlers for a specific provider.

The idea was to make the customization of individual handlers possible without affecting at all the previous user experience/the existing API, for those people not trying to do that.

So the point of this essentially is to make it possible for someone to say "I only want to run the GitHub provider", and have both of the default handlers used by default, but also if someone wants to only run the GitHub provider and customize only one of the two handlers, or both, then they can also do that (or hypothetically they could customize both handlers and then not run the GitHub provider).

Again though I hate this design and I only see it as being provisional, since my main goal was to basically include an "alpha" of new functionality in a way that doesn't at all modify existing functionality. But I think in the future another approach will be needed to support multiple levels of "granularity of customization".

krinsman: The **TL;DR** is that there is a one-to-many mapping from the "GitHub provider" to the relevant customizable handlers (contrast with the URL and localfile providers which each only have one customizable handler), so this is just a hack to implement that one-to-many mapping in a backwards compatible way, i.e. without forcing any user who's level of abstraction is only at "providers" to have to think about this one-to-many mapping if they don't care about handlers.

krinsman:
I rebased again so this now includes the Docker image fix.

parente:
@krinsman My day job is demanding my attention at the moment. I will return to reviewing these as I can. Just wanted to give you a heads up.

krinsman:
Thanks for the heads up!

parente:
I've got some time today to devote to nbviewer PRs. Picking up where we left off ...

nbviewer/app.py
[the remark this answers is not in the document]
parente: >  probably only provisional and going to have to be refactored one day
> If you want me to look into a more long-term solution for right now then I will.

Given how little time I've had and how much time I'm likely to have in the near future to devote to reviewing these PRs, I'd like to adopt the following mantra: if a code change is understandable (e.g., simple, has comments), doesn't harm the behavior of the public nbviewer instance, and adds some value for you and/or others, let's run with it.

If someone would like to spend time refactoring the code base itself to make it easier to understand, maintain, and extend, we can welcome those changes in separate PRs.

PR #845, opened 2019-07-29
Step 5

krinsman:
Refactored provider handler GET methods to make more easily extensible/customizable and to allow for inserting custom logic into the middle of the GET method, after the notebook data has been loaded, but before the notebook has been rendered.

Depends on https://github.com/jupyter/nbviewer/pull/844 ("Step 4")

PR #846, opened 2019-07-29
Step 6

krinsman:
Updated NBViewer to work with asyncio and most up-to-date version of Tornado. HTTPClient is also now finally a traitlets configurable.

Depends on https://github.com/jupyter/nbviewer/pull/845 ("Step 5")

### DIFF
--- a/nbviewer/app.py
+++ b/nbviewer/app.py
@@ -18,7 +18,6 @@ from html import escape
 from concurrent.futures import ThreadPoolExecutor, ProcessPoolExecutor
 
 from tornado import web, httpserver, ioloop, log
-from tornado.httpclient import AsyncHTTPClient
 
 import tornado.options
 from tornado.options import define, options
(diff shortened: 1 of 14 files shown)


parente:
@krinsman when you have a minute, can you rebase this change on master again. If you want to combine the Step 7 PR into it to avoid on final rebase later, feel free to do so. I think I can review the final two together.

krinsman:
@parente OK, I got around to rebasing it. I will try to rebase Step7 soon.

Also, please don't feel any rush to review these. These two PR's were the most difficult for me to write. Especially the changes to the cache in this PR -- they were quite difficult for me to figure out. Everything seems to work as before whenever I've tested it/ran it locally, but I don't feel like I can guarantee that the code is 100% equivalent to the pre-asyncio version.

parente:
> Everything seems to work as before whenever I've tested it/ran it locally, but I don't feel like I can guarantee that the code is 100% equivalent to the pre-asyncio version.

I agree this one is going to be hard to test to ensure everything is exactly the same as before the changes with the older version of tornado. That said, I think it's necessary to keep up with new versions for bug and security fixes. I'm going to QA this locally as much as I can before merging. If we find a major problem in the wild when it's deployed to nbviewer.jupyter.org, we can always roll back while we fix the issue.

parente:
Everything I tried in a local instance worked as expected. There may be handlers that still have issues. We'll figure those out over time. 

Thanks for putting in the work here, @krinsman, to move nbviewer to Tornado 6+! ❤️

PR #847, opened 2019-07-29
Step 7

krinsman:
Replaced all Tornado command line options with equivalent traitlets.config.Application ones, thereby making NBViewer a first-class Jupyter application. NBViewer is now fully configurable, with all options previously configurable via the command line now also configurable via a config file. Even location of config file is now configurable.

Added feature to write a default config file automatically, using the
command line flag --generate-config.

Updated logs to have color and look like those of other Jupyter
applications. Handlers can now call self.log instead of app_log.

Also moved /providers/url/client.py to root directory since it isn't used in /providers/url but is used by the entire application.

Depends on https://github.com/jupyter/nbviewer/pull/846 ("Step 6")

### DIFF
--- a/README.md
+++ b/README.md
@@ -181,15 +181,21 @@ Formats are ways to present notebooks to the user.
 - `script`
 
 #### Writing a new Format
-If you'd like to write a new format, open a ticket, or speak up on [gitter][]!
+If you'd like to write a new format, open a ticket, or speak up on [gitter](https://gitter.im/jupyter/nbviewer)!
 We have some work yet to do to support your next big thing in notebook
 publishing, and we'd love to hear from you.
 
-#### Config file
+## Config File and Command Line Configuration
 
-Newer versions of NBViewer will be configurable using a config file, `nbviewer_config.py`. In the directory where you run the command `python -m nbviewer` to start NBViewer, also add a file `nbviewer_config.py` which uses [the standard configuration syntax for Jupyter projects](https://traitlets.readthedocs.io/en/stable/config.html). 
+NBViewer is configurable using a config file, by default called `nbviewer_config.py`. You can modify the name and location of the config file that NBViewer looks for using the `--config-file` command line flag. (The location is always a relative path, i.e. relative to where the command `python -m nbviewer` is run, and never an absolute path.) 
 
-For example, to configure the value of a configurable `foo`, add the line `c.NBViewer.foo = 'bar'` to the `nbviewer_config.py` file located where you run `python -m nbviewer`. Again, currently very few features of NBViewer are configurable this way, but we hope to steadily increase the number of configurable characteristics of NBViewer in future releases.
+If you don't know which attributes of NBViewer you can configure using the config file, run `python -m nbviewer --generate-config` (or `python -m nbviewer --generate-config --config-file="my_custom_name.py"`) to write a default config file which has all of the configurable options commented out and set to their default values. To change a configurable option to a new value, uncomment the corresponding line and change the default value to the new value. 
+
+You can also run `python -m nbviewer --help-all` to see all of the configurable options. This is a more comprehensive version of `python -m nbviewer --help`, which gives a list of the most common ones along with flags and aliases you can use to set their values temporarily via the command line.
+
+The config file uses [the standard configuration syntax for Jupyter projects](https://traitlets.readthedocs.io/en/stable/config.html). For example, to configure the default port used to be 9000, add the line `c.NBViewer.port = 9000` to the config file. If you want to do this just once, you can also run `python -m nbviewer --NBViewer.port=9000` at the command line. (`NBViewer.port` also has the alias `port`, making it also possible to do, in this specific case, `python -m nbviewer --port=9000`. However not all configurable options have shorthand aliases like this; you can check using the outputs of `python -m nbviewer --help` and `python -m nbviewer --help-all` to see which ones do and which ones don't.)
+
+One thing this allows you to do, for example, is to write your custom implementations of any of the standard page rendering handlers included in NBViewer, e.g. by subclassing the original handlers to include custom logic along with custom output possibilities, and then have these custom handlers always loaded by default, by modifying the corresponding lines in the config file. This is effectively another way to extend NBViewer.
 
 ## Securing the Notebook Viewer
 
--- a/README.md
+++ b/README.md
@@ -43,6 +43,12 @@ $ docker run -p 8080:8080 -e 'GITHUB_OAUTH_KEY=YOURKEY' \
 
 With this configured all GitHub API requests will go to your Enterprise instance so you can view all of your internal notebooks.
 
+## Base URL
+
+If the environment variable `JUPYTERHUB_SERVICE_PREFIX` is specified, then NBViewer _always_ uses the value of this environment variable as the base URL.
+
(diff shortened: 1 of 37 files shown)


parente:
@krinsman Last one! 😲 Could you rebase on master again please so that we can retry the build?

krinsman:
@parente OK I rebased this branch.

The change from `tornado.options` to Traitlets configuration seems to have made one of the JupyterHub tests fail. This branch is currently deployed on Cori at NERSC, so it does seem to still work with JupyterHub, which is also backed up by my local testing.

So it seems like the process behind which the mock JupyterHub instance is created in the tests is somehow no longer valid with the change from `tornado.options` to Traitlets. That or the assumptions which are used for creating the mock JupyterHub in the tests. I have tried to change the tests to something more similar to the tests used in the JupyterHub repository. But I have not been able to get the JupyterHub test to pass -- it times out seemingly because the mock server does not accept any requests, perhaps because it is not created correctly. I have tried adding in print experiments and modifying the test locally to figure out what is going on, but that is the best detail I can figure out. 

The TL;DR is that unless I can figure out what is happening with that test, i.e. why the mock server isn't responding to any requests, or unless that test is removed from the test suite (which seems like a bad idea, since JupyterHub integration is a major feature of NBViewer), this PR is not going to pass the Travis build, and will fail via timeout.

parente:
Thanks for the rebase and update. I'll try to help debug when I return to reviewing.

ottomata:
Bump!  I also just ran into this problem :)

krinsman:
@ottomata To clarify, ran into which problem?

ottomata:
Sorry, I think I may have commented in the wrong comment box!

But I meant that I'm experiencing the same issue trying to run nbviewer with a more recent Tornado. Was hoping my bump would help get this reviewed and merged.  I had to install nbviewer from your fork's step7 branch to get it to work.

krinsman:
@ottomata Oh I'm glad I was able to help!

parente:
@krinsman I see the build is passing here now by skipping the failing test. I'm fine with that solution for now if you're comfortable that the functionality is not broken.

I see you've been opening various `step7_*` follow-on PRs. Are those a breakdown of this PR to debug things or additional PRs to review and merge after this one? If it's the latter, I think it's OK to stop naming them `step X` and start titling them after the feature / fix they represent.

krinsman:
@parente Oh the idea was to break up the commits into "micro-commits" so that the git diffs are easier for you to follow, although I haven't finished yet. I was looking at the diffs for this PR and I thought a lot of them were too large, or at least hard to read without knowing that most of the differences are due to e.g. some things being alphabetized.

For example, the original commit is much too large because a lot of conceptually distinct changes cause temporary feature regressions (i.e. a CLI argument is removed) before the migration to `traitlets` from `tornado.options` is completed, so I lumped all of those changes together with that migration, even though that migration itself was a huge commit, and so the entire commit is just this giant difficult-to-read monstrosity when it doesn't need to be since it is composed of many simple steps.

A pleasant side effect has been that I have figured out how to improve the implementation of some features, and another pleasant side effect has been that I think I have figured out why the problem with the `JupyterHubTestCase` occurs.

Basically the problem seems to be with the `setup_class` method of `JupyterHubTestCase`. The `wait_for_alive` line it inherits from `NBViewerTestCase` is supposed to ping the server until it gets a response and then stop. When the `prepare` method of the `Custom404Handler` isn't `await`ed (even though the `prepare` method of the parent `BaseHandler` which it uses is async), this leads to a `RunTimeWarning` which stops the `while` loop in `wait_for_alive` and allows the test to proceed. But when it _is_ `await`ed (which is probably the correct thing to do), for some reason the `302` responses given by the server in the `JupyterHubTestCase` _don't_ trigger the exit condition for the `while` loop, and so the `wait_for_alive` method in the call to `setup_class` for `JupyterHubTestCase` causes an infinite redirect loop without the `setup_class` method ever completing or any of the tests for `JupyterHubTestCase` being run.

I tried to stop the infinite redirect loop by making the `except` condition for the `while` loop more specific, e.g. by replacing `Exception` with `ConnectionError`, so that hopefully the `302` responses would no longer trigger new attempted pings of the server, but that did not seem to work. However, when I modified the `setup_class` method of `JupyterHubTestCase` to do everything the `setup_class` method of `NBViewerTestCase` does except the `wait_for_alive` part, the server still started correctly, and all of the tests for `JupyterHubTestCase` ran and passed.

Actually I wonder whether `wait_until_alive` method is necessary at all any more for the `setup_class` of any of the test cases? Would using asynchronous functions from `asyncio` instead of `tornado.gen` cause that to happen somehow? In any case it doesn't seem to cause any problems or break anything for the other test cases, so it seems like the short term solution would be to leave it as is. I also wonder if there is a way to make the `except` condition in the `while` loop more specific, so that 302 responses would no longer trigger it. But I haven't been able to figure that out yet, or what specific error/exception (if any?) the 302 responses trigger to allow the infinite redirect loop to occur when running `wait_until_alive` in the first case.

That being said all of this analysis occurred before switching from `tornado.options` to `traitlets`, so I'm not sure how much this analysis will still carry over in that case, and how much the tests might have to be refactored again to more closely resemble JupyterHub's and early NBViewer's, which is what I remember having to do last time afterwards.

krinsman:
@parente Oh, never mind about not needing to skip the JupyterHub test -- the problem seems to be different after the tests are changed to open NBViewer with `subprocess.Popen` (which seems to be necessary for some reason when switching from `tornado.options` to `traitlets`), and so the solution for not needing to skip it that worked with `tornado.options` no longer works.

If there's some way to have the logs be displayed in the Travis output, instead of redirecting them all to `os.devnull` the way the tests currently do, it might be more obvious (perhaps) how the 302 responses are causing unexpected behavior that leads to some infinite loop of some kind.

_However_, I was able to figure out how to change the other test I had skipped -- the reason why that one was consuming the `-v` flag intended for `nosetests` when the other ones weren't is because it was actually an instance of `AsyncNbviewerTestCase` and not of `NBViewerTestCase`. Since it was the only test that used `AsyncNbviewerTestCase`, and there didn't seem to be any compelling reason for it to use that instead of `NBViewerTestCase`, I changed it to use `NBViewerTestCase` (and deleted `async_base.py`), and now everything seems to work.

krinsman:
I split the 5 commits in this PR into 19 commits (and 13 PR's), as can be seen here: https://github.com/jupyter/nbviewer/pull/894

Also the implementation of some things is done better there. So after I test the branch corresponding to that to make sure that there are no feature regressions, I will probably close this PR in favor of those 13 PR's. (Or I guess just the last one, they build on each other sequentially, with the idea that the last one is roughly equivalent to this one.)

In particular please don't merge this if possible.

krinsman:
Made irrelevant by https://github.com/jupyter/nbviewer/pull/913 which was merged -- i.e. that PR, and the previous 12 (https://github.com/jupyter/nbviewer/pull/892, https://github.com/jupyter/nbviewer/pull/891, https://github.com/jupyter/nbviewer/pull/890, https://github.com/jupyter/nbviewer/pull/889, https://github.com/jupyter/nbviewer/pull/888, https://github.com/jupyter/nbviewer/pull/887, https://github.com/jupyter/nbviewer/pull/885, https://github.com/jupyter/nbviewer/pull/884, https://github.com/jupyter/nbviewer/pull/883, https://github.com/jupyter/nbviewer/pull/882, https://github.com/jupyter/nbviewer/pull/881, https://github.com/jupyter/nbviewer/pull/880) contain all of the changes that were contained in this PR, except implemented even better. So this out-of-date and will be closed.

PR #854, opened 2019-09-30
Don't print the settings in self.statsd

skritch:
:)


krinsman:
This looks like it was probably my fault -- thank you for fixing and looking into this!

parente:
Thanks!

Issue #863, opened 2019-11-18
Repo and user links do not work with GHE


ivan-gomes:
**Describe the bug**
When configured to integrate with GitHub Enterprise, links to repositories and users/orgs do not work.

Examples:
* https://ghe.mycompany.com/ivangomes
* https://ghe.mycompany.com/ivangomes/myproject 

**To Reproduce**
Enter GHE URL like https://ghe.mycompany.com/ivangomes in an nbviewer deployment configured with GHE.

**Expected behavior**
Get redirected to https://nbviewer.mycompany.com/github/ivangomes and see the navigation UI, like in https://nbviewer.jupyter.org/github/jupyter/nbviewer/tree/master/

**Additional context**
Related to #850.

krinsman:
@ivan-gomes Sorry for the slow response!

What version of NBViewer are you using? In particular, are you using the master version of NBViewer? https://github.com/jupyter/nbviewer/pull/851

ivan-gomes:
Yes I am using the master branch of nbviewer, with latest commit https://github.com/jupyter/nbviewer/commit/8a8fd298b446096dddc874da57e99dd22a2d8d97. #851 does fix some GHE compatibility like being able to look up `org/repo` or just `org`, but when the full URL is included, e.g. `https://ghe.mycompany.com/ivangomes` it does not work. This issue and corresponding PR specifically addresses that.

krinsman:
This makes more sense to me now.

To clarify, when you say `org/repo`, do you mean `<nbviewer_base_url>/github/org/repo`? `ivangomes/myrepo`, `<nbviewer_base_url>/github/ivangomes/myrepo`? Etc.?

ivan-gomes:
Assuming my repository is https://ghe.mycompany.com/ivangomes/myproject, if I typed `ivangomes/myproject` into the search box on the nbviewer home page it would redirect to `<nbviewer_base_url/github/ivangomes/myproject` properly as of https://github.com/jupyter/nbviewer/commit/8a8fd298b446096dddc874da57e99dd22a2d8d97.

However, if I instead entered the full URL, i.e. `https://ghe.mycompany.com/ivangomes/myproject` into the same search box it would instead redirect to `<nbviewer_base_url/url/https%3A%2F%2Fghe.mycompany.com%2Fivangomes%2Fmyproject` and respond 404. This issue is to address this case.

krinsman:
This makes sense to me, and seems like a really good feature enhancement!

Issue #850, opened 2019-08-28
Not compatible with Github Enterprise 


mskimm:
**Describe the bug**
 - #841 is not compatible with Github Enterprise.
 - https://github.com/jupyter/nbviewer/blob/7cf4b33afa623035ad0c321ead54501f79b54e38/nbviewer/providers/github/handlers.py#L164
     - `https://github\.com` should be fixed appropriately.

**To Reproduce**
If GITHUB_API_URL is a URL of Github Enterprise, the code in the previous section always fails.

**Expected behavior**
should work for Github Enterprise.

krinsman:
@mskimm Is this still an issue while running the master version of NBViewer?

krinsman:
@mskimm I think this was successfully closed by your merged PR, so I am tentatively closing this issue. Please let me know if you think it should be reopened!

PR #864, opened 2019-11-18
Fix repo and user links for GHE

ivan-gomes:
Closes #863.

> **Describe the bug**
> When configured to integrate with GitHub Enterprise, links to repositories and users/orgs do not work.
> 
> Examples:
> 
>     * https://ghe.mycompany.com/ivangomes
> 
>     * https://ghe.mycompany.com/ivangomes/myproject
> 
> 
> **To Reproduce**
> Enter GHE URL like https://ghe.mycompany.com/ivangomes in an nbviewer deployment configured with GHE.
> 
> **Expected behavior**
> Get redirected to https://nbviewer.mycompany.com/github/ivangomes and see the navigation UI, like in https://nbviewer.jupyter.org/github/jupyter/nbviewer/tree/master/
> 
> **Additional context**
> Related to #850.

PR #866, opened 2019-11-18
Fix Gist support with GHE

### LINKED ISSUE #865: Gist support does not work with GHE
**Describe the bug**
When configured to integrate with GitHub Enterprise, Gists do not work.

**To Reproduce**
Enter Gist URL, e.g. https://ghe.mycompany.com/gist/c3ca62e42d590f81c3a906824b2528b0, in an nbviewer configured `GITHUB_API_URL=https://ghe.mycompany.com/v3/api/`.

**Expected behavior**
The notebook in the Gist is rendered in nbviewer.

**Additional context**
Related to #850 and #863.

ivan-gomes:
Closes #865.
> **Describe the bug**
> When configured to integrate with GitHub Enterprise, Gists do not work.
> 
> **To Reproduce**
> Enter Gist URL, e.g. https://ghe.mycompany.com/gist/c3ca62e42d590f81c3a906824b2528b0, in an nbviewer configured `GITHUB_API_URL=https://ghe.mycompany.com/v3/api/`.
> 
> **Expected behavior**
> The notebook in the Gist is rendered in nbviewer.
> 
> **Additional context**
> Related to #850 and #863.

### DIFF
--- a/nbviewer/providers/gist/handlers.py
+++ b/nbviewer/providers/gist/handlers.py
@@ -18,6 +18,7 @@ from ...utils import (
     clean_filename,
     quote,
     response_text,
+    url_path_join,
 )
 
 from ..github.handlers import GithubClientMixin
@@ -97,8 +98,11 @@ class UserGistsHandler(GistClientMixin, BaseHandler):
                     notebooks=notebooks,
                     description=gist['description'] or '',
                 ))
-        gist_url = os.environ.get('GIST_URL', 'https://gist.github.com/')
-        provider_url = gist_url + u"{user}".format(user=user)
+        if self.github_url == 'https://github.com/':
+            gist_base_url = 'https://gist.github.com/'
+        else:
+            gist_base_url = url_path_join(self.github_url, 'gist/')
+        provider_url = url_path_join(gist_base_url, u"{user}".format(user=user))
         html = self.render_usergists_template(entries=entries, user=user, provider_url=provider_url, 
                                               prev_url=prev_url, next_url=next_url, **namespace
 
@@ -158,12 +162,15 @@ class GistHandler(GistClientMixin, RenderingHandler):
                 e['class'] = 'fa-book'
                 ipynbs.append(e)
             else:
-                gist_url = os.environ.get('GIST_URL', 'https://gist.github.com/')
-                provider_url = gist_url + u"{user}/{gist_id}#file-{clean_name}".format(
+                if self.github_url == 'https://github.com/':
+                    gist_base_url = 'https://gist.github.com/'
+                else:
+                    gist_base_url = url_path_join(self.github_url, 'gist/')
+                provider_url = url_path_join(gist_base_url, u"{user}/{gist_id}#file-{clean_name}".format(
                     user=user,
                     gist_id=gist_id,
                     clean_name=clean_filename(file['filename']),
-                )
+                ))
                 e['url'] = provider_url
                 e['class'] = 'fa-share'
                 others.append(e)
@@ -316,11 +323,11 @@ def uri_rewrites(rewrites=[]):
             u'/{1}'),
     ] 
     # github enterprise
-    if os.environ.get('GIST_URL', '') != '':
-        gist_url = os.environ.get('GIST_URL')
+    if os.environ.get('GITHUB_API_URL', '') != '':
+        gist_base_url = url_path_join(os.environ.get('GITHUB_API_URL').split('/api/v3')[0], 'gist/')
         gist_rewrites.extend([
-            # embedded in URL
-            (r'^' + gist_url + r'([^\/]+/)?([a-f0-9]+)/?$',
+            # Fetching the Gist ID which is embedded in the URL, but with a different base URL
+            (r'^' + gist_base_url + r'([^\/]+/)?([a-f0-9]+)/?$',
                 u'/{1}'),
         ])
 
--- a/nbviewer/providers/gist/handlers.py
+++ b/nbviewer/providers/gist/handlers.py
@@ -1,6 +1,7 @@
 # Copyright (c) Jupyter Development Team.
 # Distributed under the terms of the Modified BSD License.
 
+import os
 import json
 
 from tornado import web, gen
@@ -79,9 +80,11 @@ class UserGistsHandler(GistClientMixin, BaseHandler):
                     notebooks=notebooks,
                     description=gist['description'] or '',
                 ))
-        provider_url = u"https://gist.github.com/{user}".format(user=user)
+        gist_url = os.environ.get('GIST_URL', 'https://gist.github.com/')
+        provider_url = gist_url + u"{user}".format(user=user)
         html = self.render_usergists_template(entries=entries, user=user, provider_url=provider_url, 
                                               prev_url=prev_url, next_url=next_url, **namespace
+
         )
         yield self.cache_and_finish(html)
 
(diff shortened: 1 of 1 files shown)


krinsman:
Ditto the comments on https://github.com/jupyter/nbviewer/pull/864

ivan-gomes:
> Can you post some screenshots of the testing for this which you've done locally?

Unfortunately I can't do that. If there is another way to demonstrate it that you can recommend, I would be happy to try.

> Were there still any issues after you ran this modified version of NBViewer?

Other GHE related issues are https://github.com/jupyter/nbviewer/issues/863 and https://github.com/jupyter/nbviewer/issues/865. These three are all GHE related issues that we identified, which I believe cover all the features.

krinsman:
The questions about this I have are similar to the ones for #864 but

Why do the strings have to be raw strings?

What is the purpose of the `GIST_URL` environment variable? 

I agree that the way this is written it doesn't seem to cause any harm, since the default behavior is the same as the previous behavior in master. 

Would the value of `GIST_URL` be any different from the value of `GITHUB_API_URL` for a GitHub Enterprise user? 

What is the expected relationship between the values of `GITHUB_API_URL` and `GIST_URL`?

Would it make sense to change the name of the environment variable to be `GIST_GITHUB_API_URL` or something like that?  

Would it also be possible to add one or two lines to the README (unfortunately NBViewer doesn't have a standalone `docs` folder yet) explaining what the environment variable does and what its value should be set to for anyone who wants to use it? (And would it be used only by people using GitHub Enterprise?)

What is the section under the comment `#embedded in URL` doing?

Also I wonder if `GIST_URL` and `GITHUB_API_URL` are typically expected to have the same value, if it would make more sense to have that be a setting for the NBViewer application which is then passed to the `BaseHandler` and then could be accessed that way?

ivan-gomes:
> Why do the strings have to be raw strings?

The motivation was to directly parallel the `url_rewrites` for the non-GHE cases right above: https://github.com/jupyter/nbviewer/blob/0ddbaf1fcc915b6bb84dff56b8406bb66c65d387/nbviewer/providers/gist/handlers.py#L315-L316

> What is the purpose of the `GIST_URL` environment variable?

The `GIST_URL` is the base path for Gists, which in the public context is https://gist.github.com/ (default) but for GHE is different, e.g. https://ghe.mycompany.com/gist. I do not know whether the path is configurable, so I thought it would be safest to make it an environment variable instead of assuming `gist/` derived from `GITHUB_API_URL`. I am fine with `GITHUB_API_URL - 'api/v3/' + 'gist/'` if preferred.

> I agree that the way this is written it doesn't seem to cause any harm, since the default behavior is the same as the previous behavior in master.
> 
> Would the value of `GIST_URL` be any different from the value of `GITHUB_API_URL` for a GitHub Enterprise user?

Addressed above. Examples: `GIST_URL=https://ghe.mycompany.com/gist/` and `GITHUB_API_URL=https://ghe.mycompany.com/api/v3/`.

> What is the expected relationship between the values of `GITHUB_API_URL` and `GIST_URL`?

Addressed above. In my sample size of one, `GIST_URL` = `GITHUB_API_URL` - 'api/v3/' + 'gist/'.

> Would it make sense to change the name of the environment variable to be `GIST_GITHUB_API_URL` or something like that?

It's not quite an API url, more a base path/url. I think `GIST_BASE_URL` is a better alternative.

> Would it also be possible to add one or two lines to the README (unfortunately NBViewer doesn't have a standalone `docs` folder yet) explaining what the environment variable does and what its value should be set to for anyone who wants to use it? (And would it be used only by people using GitHub Enterprise?)

Yes, I will add to the README soon. And yes, the only use case I see is with GHE.

> What is the section under the comment `#embedded in URL` doing?

It is redirecting a user input of `https://ghe.mycompany.com/gist/XYZ` to `<nbviewer_base_path/gist/XYZ`. Directly parallels https://github.com/jupyter/nbviewer/blob/0ddbaf1fcc915b6bb84dff56b8406bb66c65d387/nbviewer/providers/gist/handlers.py#L315-L316.

> Also I wonder if `GIST_URL` and `GITHUB_API_URL` are typically expected to have the same value, if it would make more sense to have that be a setting for the NBViewer application which is then passed to the `BaseHandler` and then could be accessed that way?

Addressed above.

krinsman:
OK this looks good to me, just like the other PR's. 

Would it be possible to change the name of the `gist_url` variable to `gist_base_url` or something like that? `GistClientMixin` doesn't have an attribute named `gist_url` yet, but I am worried about people looking at this several years from now without context finding `gist_url` to be too broad or assuming it does other things which it does not do.

I agree with your suggestion to change the value of the `gist_base_url`/`gist_url` to `GITHUB_API_URL` - `api/v3/` + `gist/`. As you pointed out, there is the extra price of adding extra assumptions to the code which might become invalid over time, but (1) it would decrease the amount of manual configuration the end-user needs to perform, and (2) I could imagine some users being confused about exactly they need to set the value of the environment variable to be. Making it as easy as possible for users seems like it should be a priority inasmuch as the features of NBViewer are not yet thoroughly documented.

ivan-gomes:
Will try to address those and tag you when it could use your eyeballs.

ivan-gomes:
@krinsman how's that look?

krinsman:
This looks really good to me! Easy to read and understand.

Regarding the checks for whether GitHub Enterprise is being used, would it be possible to use the same
```  if os.environ.get('GITHUB_API_URL', '') != '':```
check you have in line 327 throughout?

I know I'm overthinking this, but because of the stuff with the slashes and computers being overly literal (i.e. stupid), I'm afraid that the earlier checks of the form:
``` if self.github_url == 'https://github.com/':```
could lead to unintended behavior?

On the other hand, the earlier checks do have the benefit of making full use of the `github_url` property of `GitHubClientMixin`/`GistClientMixin`. 

I don't know if there's something in urllib which would cover all possible variants of `http://github.com`, `https://github.com`, `http://github.com/`, `https://github.com/`, etc., but maybe that could be an alternative to re-computing the value of `self.github_url` directly via the environment variable?

You're right, `https://github.com` is effectively the default value for `github_url`
https://github.com/jupyter/nbviewer/blob/master/nbviewer/providers/github/handlers.py#L67
so I guess it shouldn't lead to unexpected behavior.

It still seems kind of brittle though, since really what we want to do in all cases is check whether the value of `GitHubClientMixin`'s `github_url` property is equal to its default or not (regardless of what that default value is, which maybe could be changed in future PR's and then unintentionally break things), and if not assume that GitHub Enterprise is being used. But there doesn't seem to be a clean interface for doing that.

Maybe some sort of observer pattern using traitlets and `observe`/`validate`? 
https://traitlets.readthedocs.io/en/stable/using_traitlets.html

Honestly I'm really not sure at all either way, and am looking/asking for your opinion on all of this as much as I'm doing anything else. Sorry again for the ridiculous nitpicks.

krinsman:
At the very least maybe adding comments to the code making explicit that

> what we want to do in all cases is check whether the value of GitHubClientMixin's github_url property is equal to its default or not (regardless of what that default value is, which maybe could be changed in future PR's and then unintentionally break things), and if not assume that GitHub Enterprise is being used

so that if something gets broken in a future PR because someone changes the default value of `github_url`, it will take them less time to figure out what happened.

ivan-gomes:
> the earlier checks do have the benefit of making full use of the `github_url` property of `GitHubClientMixin`/`GistClientMixin`.

That was exactly the intent. The Mixin isn't available for `url_rewrite` which is why it's not used there.

However, I agree with your points and think a clean solution could be to introduce a `gist_base_url` def to `GistClientMixIn` that mirrors https://github.com/jupyter/nbviewer/blob/8a8fd298b446096dddc874da57e99dd22a2d8d97/nbviewer/providers/github/handlers.py#L62-L72 and is re-used in both places. That way there is only one place to maintain the default `github_url` matching condition and default `gist_base_url` value. It won't address the `url_rewrite` case, but there's not much to be done there without a heavier refactor.

krinsman:
OK, well maybe a comment above line 101 pointing out that `https://github.com/` is supposed to be the default value of `self.github_url`, and the point is that if it isn't equal to the default the assumption is because Github Enterprise is available would still be helpful.

Regardless the substance is already good as is, so I will request @parente review before merging. Again I think the comment clarifying that would still be helpful though, in case a slash gets added or removed to the default value sometime in the future and everything breaks and no one knows why.

PR #868, opened 2019-11-18
Fix raw download for GHE

### LINKED ISSUE #867: Raw download link does not work with GHE
**Describe the bug**
When configured to integrate with GitHub Enterprise, raw download links for notebooks do not work.

**To Reproduce**
With a notebook from GHE open, e.g. https://nbviewer.mycompany.com/github/ivangomes/myproject/index.ipynb, in an nbviewer configured GITHUB_API_URL=https://ghe.mycompany.com/v3/api/, click the download button on the top right and get a 404.

**Expected behavior**
File download from GHE starts.

**Additional context**
Related to #850, #863, #865.

ivan-gomes:
Closes #867.
> **Describe the bug**
> When configured to integrate with GitHub Enterprise, raw download links for notebooks do not work.
> 
> **To Reproduce**
> With a notebook from GHE open, e.g. https://nbviewer.mycompany.com/github/ivangomes/myproject/index.ipynb, in an nbviewer configured GITHUB_API_URL=https://ghe.mycompany.com/v3/api/, click the download button on the top right and get a 404.
> 
> **Expected behavior**
> File download from GHE starts.
> 
> **Additional context**
> Related to #850, #863, #865.

### DIFF
--- a/nbviewer/providers/github/handlers.py
+++ b/nbviewer/providers/github/handlers.py
@@ -262,8 +262,12 @@ class GitHubBlobHandler(GithubClientMixin, RenderingHandler):
     @cached
     @gen.coroutine
     def get(self, user, repo, ref, path):
-        raw_url = u"https://raw.githubusercontent.com/{user}/{repo}/{ref}/{path}".format(
-            user=user, repo=repo, ref=ref, path=quote(path)
+        if os.environ.get('GITHUB_API_URL', '') == '':
+            raw_pattern = u"https://raw.githubusercontent.com/{user}/{repo}/{ref}/{path}"
+        else: #Github Enterprise has a different URL pattern for accessing raw files
+            raw_pattern = u"{github_url}{user}/{repo}/raw/{ref}/{path}"
+        raw_url = raw_pattern.format(
+            user=user, repo=repo, ref=ref, path=quote(path), github_url=self.github_url
         )
         blob_url = u"{github_url}{user}/{repo}/blob/{ref}/{path}".format(
             user=user, repo=repo, ref=ref, path=quote(path), github_url=_github_url()
--- a/nbviewer/providers/github/handlers.py
+++ b/nbviewer/providers/github/handlers.py
@@ -29,6 +29,7 @@ from ...utils import (
     base64_decode,
     quote,
     response_text,
+    url_path_join,
 )
 
 from .client import AsyncGitHubClient
@@ -291,12 +292,13 @@ class GitHubBlobHandler(GithubClientMixin, RenderingHandler):
     @gen.coroutine
     def get_notebook_data(self, user, repo, ref, path):
         if os.environ.get('GITHUB_API_URL', '') == '':
-            raw_pattern = u"https://raw.githubusercontent.com/{user}/{repo}/{ref}/{path}"
+            raw_url = u"https://raw.githubusercontent.com/{user}/{repo}/{ref}/{path}".format(
+                user=user, repo=repo, ref=ref, path=quote(path)
+            )
         else: #Github Enterprise has a different URL pattern for accessing raw files
-            raw_pattern = u"{github_url}{user}/{repo}/raw/{ref}/{path}"
-        raw_url = raw_pattern.format(
-            user=user, repo=repo, ref=ref, path=quote(path), github_url=self.github_url
-        )
+            raw_url = url_path_join(
+                self.github_url, user, repo, 'raw', ref, quote(path)
+            )
         blob_url = u"{github_url}{user}/{repo}/blob/{ref}/{path}".format(
             user=user, repo=repo, ref=ref, path=quote(path), github_url=self.github_url
         )


krinsman:
Thank you so much for submitting fixes for all of these! We really appreciate it!

Ditto the comments on #864

ivan-gomes:
> Can you post some screenshots of the testing for this which you've done locally?

Unfortunately I can't do that. If there is another way to demonstrate it that you can recommend, I would be happy to try.

> Were there still any issues after you ran this modified version of NBViewer?

Other GHE related issues are https://github.com/jupyter/nbviewer/issues/863 and https://github.com/jupyter/nbviewer/issues/865. These three are all GHE related issues that we identified, which I believe cover all the features.

krinsman:
OK, these changes look good, and again don't change the default behavior for users not using GitHub Enterprise, which is definitely a plus.

Would it be possible to maybe import `url_path_join` from `nbviewer.utils` (so I guess in this file it would be `..utils` or `...utils`, I don't remember which) and concatenate the parts that way?

The current `.format` approach where some parts have `/`'s inserted between them and some don't seems kind of confusing and fragile. I don't think this is your fault, `url_path_join` or something like it probably should be used more often by default in the NBViewer code, and it seems like you've found a good opportunity/place to utilize it more.

ivan-gomes:
Makes sense to me. I will try to add that soon.

krinsman:
Great! Other than that, I would also recommend that this be merged into master.

ivan-gomes:
@krinsman how's that look?

krinsman:
This looks good to me!

As a user of GitHub Enterprise with NBViewer, can you explain to me what the difference is between the `GITHUB_URL` and `GITHUB_API_URL` environment variables?

The `github_url` attribute of the `GitHubClientMixin` seems to be based on the `GITHUB_URL` environment variable. This surprised/confused me, because I thought that the environment variable was called `GITHUB_API_URL`. (Of course part of the problem is that NBViewer's ability to work with GitHub enterprise is not as well documented as it should be, but...)

Are they supposed to represent the same thing? If so, do you think it would make sense to expand the scope of this PR by changing the lines of code like
https://github.com/jupyter/nbviewer/pull/868/files#diff-f2de296d9ff85c0b7d3f77f66cbaf291L64
which refer to a `GITHUB_URL` environment variable and replace them with references to a `GITHUB_API_URL` environment variable?

I understand that this discrepancy in environment variable names isn't a result of these PR's at all by any means. This seems to me like it could be a good time to fix this discrepancy and simplify NBViewer's interface, making it easier for users.

If there is a difference between the two environment variables, would you consider adding one or two lines to the `README.md` explaining what the distinction is? In the future we should probably have a separate `docs` folder building a whole `readthedocs.io` website using Sphinx and restructuredText, but right now all we have is the `README.md`.

https://github.com/jupyter/nbviewer#github-enterprise

Sorry for all of these nitpicks -- I think the substance of these PR's is great. 

They demonstrate a really clear and strong understanding of how to use NBViewer with GitHub enterprise. The goal is to refine them further so that the GitHub enterprise features are sufficiently easy to use out-of-the-box for everyone that no one else has to selflessly spend as much time as you did making PR's to fix them.

ivan-gomes:
The short version is: only `GITHUB_API_URL` is necessary for GitHub Enterprise integration, and I do not know any reason that the `GITHUB_URL` environment variable needs to exist.

To go into more details:
The `github_url` attribute is used to parse the URLs that the user inputs, whereas `github_api_url` is GitHub's API base path and used to make requests to the API.

In turn, `GITHUB_URL` and `GITHUB_API_URL` are two separate environment variables.

However, `GITHUB_API_URL` is the *only* environment variable that should be used for GitHub Enterprise deployments and in fact it is the only one documented. This is because for GitHub's REST API v3, the `github_url` attribute can be derived from the `github_api_url` attribute (and it is - see https://github.com/jupyter/nbviewer/blob/8a8fd298b446096dddc874da57e99dd22a2d8d97/nbviewer/providers/github/handlers.py#L71).

My guess is that `GITHUB_URL` is a solution by explicit specification to some obscure edge cases (perhaps pre-v3) where one GitHub Enterprise deployment would have a different API endpoint than the classic form of `http(s)://ghe.mycompany.com/api/v3/`. However, I do not know if this is even possible/necessary anymore.

---

> If so, do you think it would make sense to expand the scope of this PR by changing the lines of code [...] which refer to a `GITHUB_URL` environment variable and replace them with references to a `GITHUB_API_URL` environment variable?

I am game to implement and test such a change against GHE, but one of the selfish motivations of these PRs is to not have to maintain a fork with these patches for longer than necessary. I worry that increasing the scope may delay that, and if so I would rather do it as a separate PR. I would also like to establish consensus that our understanding of `GITHUB_URL`'s purpose is correct, which @parente or @mskimm could likely speak to.

krinsman:
Ok yes that is fair and reasonable, making it a separate PR makes sense.

Alright, I will request that @parente review this before merging, but it looks good to me!

PR #870, opened 2019-12-10
Fix #869 from cgi import escape fails under python 3.8; import from html instead

### LINKED ISSUE #869: from cgi import escape fails under python 3.8; import from html instead
**Describe the bug**
Importing 'escape' from cgi fails under python3.8 
at line 15 of providers/base.py

**To Reproduce**
>>> from cgi import escape

**Expected behavior**
Successful import

**Screenshots**
>>> from cgi import escape
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ImportError: cannot import name 'escape' from 'cgi' 

**Resolution**
replace line 15 of providers/base.py with
`from html import escape`

MiddleFork:
replace line 15 of providers/base.py with
`from html import escape`

### DIFF
--- a/nbviewer/providers/base.py
+++ b/nbviewer/providers/base.py
@@ -12,7 +12,7 @@ import time
 import statsd
 import asyncio
 
-from cgi import escape
+from html import escape
 from contextlib import contextmanager
 from datetime import datetime
 


krinsman:
Would it be possible to change this to:

```
# Python 3.8 Compatibility
try:
    from cgi import escape
except ImportError:
    from html import escape
```

to ensure backwards-compatibility?

Compare e.g. https://github.com/jupyter/nbviewer/blob/master/nbviewer/app.py#L41

krinsman:
I agree that it probably makes no difference, but the default behaviors of the two functions are _slightly_ different, also if changing it to `html`'s `escape` causes unexpected behavior, having the whole try-except block might make the source of the unexpected behavior easier to track down.

krinsman:
Again @MiddleFork I would approve these changes if the try-except block for version backwards-compatibility were added. I only have triage access though so I can't modify your PR myself to incorporate nor can I merge the PR myself. But I definitely think this is good and it or at least something very like it should be merged!

krinsman:
An alternative @parente could be to just merge this directly as is. 

And maybe later PR's can add the exception handling. It seems like the change is probably backwards compatible with previous versions of Python, but I thought the exception handling might provide stronger guarantees that nothing will break, but in hindsight it might be overkill.

PR #882, opened 2020-01-28
alphabetize BaseHandler properties

krinsman:
Alphabetize BaseHandler properties -- this makes it much easier to find and modify relevant parts of the code, at least in my experience.

### DIFF
--- a/nbviewer/providers/base.py
+++ b/nbviewer/providers/base.py
@@ -130,105 +130,50 @@ class BaseHandler(web.RequestHandler):
                 raise ex
 
     # Properties
-    @property
-    def pending(self):
-        return self.settings.setdefault('pending', {})
 
     @property
-    def formats(self):
-        return self.settings['formats']
-
-    @property
-    def providers(self):
-        return self.settings['providers']
-
-    @property
-    def default_format(self):
-        return self.settings['default_format']
-
-    @property
-    def config(self):
-        return self.settings['config']
-
-    @property
-    def client(self):
-        return self.settings['client']
-
-    @property
-    def index(self):
-        return self.settings['index']
-
-    @property
-    def cache(self):
-        return self.settings['cache']
-
-    @property
-    def cache_expiry_min(self):
-        return self.settings.setdefault('cache_expiry_min', 60)
-
-    @property
-    def cache_expiry_max(self):
-        return self.settings.setdefault('cache_expiry_max', 120)
-
-    @property
-    def rate_limiter(self):
-        return self.settings['rate_limiter']
-
-    @property
-    def pool(self):
-        return self.settings['pool']
-
-    @property
-    def max_cache_uris(self):
-        return self.settings.setdefault('max_cache_uris', set())
-
-    @property
-    def frontpage_setup(self):
-        return self.settings['frontpage_setup']
-
-    @property
-    def mathjax_url(self):
-        return self.settings['mathjax_url']
-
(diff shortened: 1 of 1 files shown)


nbviewer/providers/base.py
```diff
@@ -442,27 +464,6 @@ def write_error(self, status_code, **kwargs):
     # response caching
     #---------------------------------------------------------------
 
-    @property
```
minrk: I think it's better to leave the cache-related things here in the cache-related section, so that related items are together. Can we leave most, if not all of these alone, or at least the ones that have logical groupings?

krinsman: I agree with your reasoning and will make those changes soon, when the time comes up.

I was originally waffling about whether or not to organize them that way, since I couldn't decide for myself what seemed most "consistent".

minrk: It's hard to choose! I think logical groupings are useful, but they can make it hard to intuitively scroll to your destination since you have to know something's in a group, as I'm sure you've found. It does make it easier to find nearby relevant things, though, so it's a trade off.

I don't think the bulk explicit property->settings key is super sustainable. Maybe overriding `__getattr__` to look in settings for most things by default is better?

krinsman: I wouldn't have been able to express it in those terms/so eloquently myself, but yes that was exactly the choice I was confused about. With the logical groupings there also seemed to be the issue that some of the ones that are currently in the giant list of properties probably should belong to their own groups, or some of the existing groups, and then which ones belong to existing groups? What would the new groups be? Should there be a formal layer of abstraction encapsulating those groups, and if so what would be the best choice?

Anyway I am looking into putting back the ones which were originally in logical groups and leaving it at that for now.

I also 100% agree that the bulk explicit property -> settings key isn't very sustainable. My only concern with overriding `__getattr__` is that might lead to an ignorant/naive person like me in the future looking at the code and seeing the `__getattr__` calls and assuming that it's the built-in version, and then writing code based on that assumption, and then the code doesn't work and then they're confused and upset and don't understand why it isn't working the way it's supposed to, yadda, yadda.

Although I suppose maybe something which does a `try`/`except` block, where it first tries the Python built-in version, and then if it doesn't work then it looks in settings, could work and avoid the above issue? I agree that having to fish everything out of the Tornado handler's settings attribute isn't actually efficient.

krinsman:
OK, this seems to fail for me only because more than 60 GitHub API requests are being made, although that doesn't seem to be an issue for the tests run in Jupyter master.

So, _provisionally_, I will merge this, to see if the tests pass when run for the Jupyter repo (as opposed to the `krinsman` one) and if it still fails, then I will revert the merge.

This is because I believe I made the requested changes.

krinsman:
The GitHub tests which pass after the ones that fail seem to pass only because they don't require calls to the GitHub API: https://github.com/jupyter/nbviewer/blob/master/nbviewer/providers/github/tests/test_handlers.py

I don't know if the GitHub authorization is better for the Jupyter test suite such that all of the API requests can be made without running out.

**Update:** Yes, it seems to be the case that the tests are failing only for me because my Travis enviornment doesn't have better than default GitHub authorization.

https://travis-ci.org/jupyter/nbviewer/builds/654621416?utm_source=github_status&utm_medium=notification

Anyway it is nice being able to run all of the tests now and not needing to skip the GitHub ones.

PR #883, opened 2020-01-28
alphabetize NBViewer settings

krinsman:
Alphabetize nbviewer application attributes, properties, settings, and options. This PR will make the git diff for the changing of tornado options to traitlets easier to read.


krinsman:
This should not actually change the behavior of the code at all. Also, the only reason the 3.6 build failed seems to be because of GitHub rate limit issues. The 3.5 build passed. So I will tentatively merge this, and revert if it breaks anything.

PR #885, opened 2020-01-29
Step7_6 (custom static path now behaves same as custom template path)

krinsman:
Changes to static file handling to facilitate customization. In particular, custom static paths now behave the same way that custom template paths do.

This involved changing the `static_url` method of `BaseHandler` to be different from the Tornado default. Moreover, the custom static file handler from notebook (instead of the default Tornado static file handler) is used (since it supports loading static files from multiple directories).

### DIFF
--- a/nbviewer/app.py
+++ b/nbviewer/app.py
@@ -43,7 +43,7 @@ try: # Python 3.8
 except ImportError:
     from .utils import cached_property
 
-from notebook.base.handlers import FileFindHandler as StaticFileHandler
+from jupyter_server.base.handlers import FileFindHandler as StaticFileHandler
 
 #-----------------------------------------------------------------------------
 # Code
--- a/nbviewer/app.py
+++ b/nbviewer/app.py
@@ -43,6 +43,8 @@ try: # Python 3.8
 except ImportError:
     from .utils import cached_property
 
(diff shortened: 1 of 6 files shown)


krinsman:
This again seems to be fine with respect to Travis builds except for my not having authorization in my personal account for a higher GitHub API rate limit.

krinsman:
@parente This PR would add `notebook` as a dependency. Is that OK? 

That's arguably a heavy dependency. And it's only needed for `FileFindHandler`, so I could modify this PR to copy-paste the code from there into a new file in the NBViewer repository. 

But that would prevent bug fixes merged into Notebook's version from automatically being reflected here. So I am not sure which is the better option.

krinsman:
I guess you made a maintainer so that I could make small decisions like this by myself.

Provisionally we can use `jupyter_server` for sourcing the `FileFindHandler`, but if it turns out to be too large of a dependency (say because of downloading `ipykernel` or `pyzmq`) in the future, we can switch to the copy-paste a version for `nbviewer`.

An interesting future project might be to say if `BaseHandler` could be sub-classed from `JupyterHandler` or one of the other handlers defined in `jupyter_server`. It probably can't, but if it could, that might be helpful/a good idea for decreasing NBViewer's technical debt and thus long-term maintainability.

Anyway, merging for now *fingers crossed*

PR #887, opened 2020-01-30
Step7_7 (prepare base_url for traitlets by making it "private")

krinsman:
For traitlets, the CLI specified value _always_ overwrites the available default, whereas for `base_url` we want the opposite behavior, i.e. for the default (an environment variable) to _always_ overwrite the CLI specified value, if the default is available.

Therefore, the solution I thought of was to make the "true" value of base url private, and to have the value specified on the CLI to just be the back-up value in case the `JUPYTERHUB_SERVICE_PREFIX` environment variable isn't available. 

Come to think of it, this behavior should probably be explicitly documented somewhere, since it might be counterintuitive to someone expecting typical traitlets behavior.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,12 @@ $ docker run -p 8080:8080 -e 'GITHUB_OAUTH_KEY=YOURKEY' \
 
 With this configured all GitHub API requests will go to your Enterprise instance so you can view all of your internal notebooks.
 
+## Base URL
+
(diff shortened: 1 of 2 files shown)


krinsman:
This is passing tests even in my GitHub API rate-deficient fork, so I will merge it for now. These changes are very minimal, and are just a small part of the Step7 PR which is already running without issues at NERSC, so I am at least provisionally confident that there shouldn't be any breaking changes.

PR #888, opened 2020-01-30
Step7_8 (Add ability to write default config file)

krinsman:
Note that `options.answer_yes`, `options.generate_config`, and `options.config_file` will all be turned into traitlets upon the transfer from `tornado.options` to `traitlets`. Writing them as `tornado.options` for now was just a provisional step so I could write this commit before the `tornado.options -> traitlets` commit, which will necessarily have a large diff. So splitting as much off from that diff as possible seems optimal.

### DIFF
--- a/README.md
+++ b/README.md
@@ -193,9 +193,13 @@ publishing, and we'd love to hear from you.
 
 #### Config file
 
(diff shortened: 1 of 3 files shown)


krinsman:
The Python 3.6 build failed only because of GitHub API rate limit issues on my fork. It even passed the new write_config_file (admittedly copied and lightly modified from JupyterHub, just like the `write_config_file` method of the application). So it doesn't seem like there are any breaking changes, and one wouldn't expect any, since none of this code actually touches the main functioning of the server.

So I will tentatively also merge this for now.

PR #889, opened 2020-01-30
Step7_9 (Create function calculating default host and port)

krinsman:
Prepare host and port for traitlets refactor and accompanying deletion of init_options


krinsman:
This passed tests even on my GitHub API Rate deficient fork, and is basically a very minor change anyway, so I feel relatively confident merging this provisionally.

The real monster change comes with Step7_13, the jump to Traitlets from tornado.options. (Well the logging change is also fairly major.) But this one is not much of a cause for concern for me so I will merge it now provisionally.

PR #890, opened 2020-01-30
Step7_10 (Make configure_formats method of NBViewer)

krinsman:
Make configure_formats a method of `NBViewer` to prepare for deletion of the options object with transition from `tornado.options` to `traitlets`.

In particular, the `configure_formats` function as written in `formats.py` depended on the `options` argument created by `tornado.options` in order to access the value of "`options.processes`".

After the transition from `tornado.options` to `traitlets`, any reference to `options.processes` will need to be replaced by a reference to `self.processes` inside of the `NBViewer` class. Therefore in order to access that value, `configure_formats` needs to be a method of `NBViewer`. 

(Or somehow it would be necessary to pass the entire application to `configure_formats`, if it were left as its own standalone function in `formats.py`, and then access `nbviewer.processes` if `nbviewer = NBViewer()`, but that seems like overkill and to be honest I don't think I would know how to do it anyway. Whereas I could figure out how to do this.)

This involves a _temporary_ feature regression inasmuch as changing `options.processes` to a traitlet `self.processes` before the `tornado.options -> traitlets` transition begins/is completed means that the only way to configure `processes` will _temporarily_ be via `nbviewer_config.py`, and impossible via the command line (which is currently parsed by `tornado.options`).

(I tried to ease the transition by having both `tornado.options` and `traitlets` parse the command line simultaneously when I first wrote this, but to make a long story short basically everything broke and it was necessary to change everything from `tornado.options` to `traitlets` at the same time, which is what I initially did, or accept _temporary_ regressions, like I'm doing now, where some features that used to be configurable via the CLI become _temporarily_ configurable only via a config file. This is a main, if not the main, reason why I originally insisted on trying to keep Step7 as a single PR for such a long time, even though it involves numerous changed lines of code.)

However, once the transition to `traitlets` is completed, this will _not_ be a feature regression, and will instead be a feature _enhancement_, since then `processes` will be configurable via _both_ the CLI and via a config file.

### DIFF
--- a/nbviewer/app.py
+++ b/nbviewer/app.py
@@ -24,13 +24,15 @@ from tornado.options import define, options
 
 from jinja2 import Environment, FileSystemLoader
 
-from traitlets import Any, Dict, Set, Unicode, default
+from traitlets import Any, Bool, Dict, Int, List, Set, Unicode, default
 from traitlets.config import Application
 
 from .handlers import init_handlers
 from .cache import DummyAsyncCache, AsyncMultipartMemcache, MockCache, pylibmc
 from .index import NoSearch
-from .formats import configure_formats
+from .formats import default_formats
+from nbconvert.exporters.export import exporter_map
+
 from .providers import default_providers, default_rewrites
 from .providers.url.client import NBViewerAsyncHTTPClient as HTTPClientClass
 from .ratelimit import RateLimiter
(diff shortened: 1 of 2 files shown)


krinsman:
Again, this is a minor change, and the only failing tests on my fork are due to GitHub API rate limit issues, which this PR shouldn't actually affect.

So I will provisionally merge this.

PR #891, opened 2020-01-31
Step7_11 (Unify and update logging)

krinsman:
Unify all logging to a central log which is an attribute of the NBViewer class. Update format of logs to resemble that of other Jupyter applications, and to be more informative.

This involves a _temporary_ regression, since in order to use the `log_level` attribute of  `traitlets.config.Application`, and set it to `DEBUG` level via a command line flag, it is necessary to activate `traitlets` parsing of command line arguments, which can't be initiated until the migration of all `tornado.options` to `traitlets` is complete. I believe this is the reason why I previously had the changes to the logging and the entire migration to `traitlets` in one massive commit: I wanted to avoid any commits with even temporary feature regressions.

Again however this will involve zero loss of features, and the addition of several new features, after all `tornado.options` are replaced with `traitlets` and command line parsing is handed over to `traitlets` from `tornado.options`.

### DIFF
--- a/nbviewer/app.py
+++ b/nbviewer/app.py
@@ -17,7 +17,9 @@ from html import escape
 
 from concurrent.futures import ThreadPoolExecutor, ProcessPoolExecutor
 
-from tornado import web, httpserver, ioloop, log
+from tornado import web, httpserver, ioloop
+from tornado.log import access_log, app_log, LogFormatter
+from tornado.curl_httpclient import curl_log
 
 import tornado.options
 from tornado.options import define, options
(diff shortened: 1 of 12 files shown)


krinsman:
This has a fair number of changes to lines of code, but in terms of substance, this is also a fairly minimal change. Patched one or two bugs with the GitHub providers too (namely the code expected a tuple when sometimes a `None` could be returned instead, so added case handling). The tests are passing even on my GitHub rate limit deficient fork, the logs are behaving as expected now on Travis, and these are the same changes that are already running now on NERSC with no ill effect. So I feel fairly confident merging this PR, at least provisionally.

PR #892, opened 2020-02-01
Step7_12

krinsman:
Moved client file to base since it's used by the NBViewer application, not specifically the URL provider. Removed some unused imports, updated copyright statements, and general tidying before tornado.options -> traitlets migration


krinsman:
This is passing all tests on my fork https://travis-ci.com/krinsman/nbviewer/builds/150768011
and doesn't actually do anything besides rename a file, update an internal import to be aware of that file name, and update the copyright comments. So I'm fairly confident that merging this will cause no changes.

PR #894, opened 2020-02-04
Step7_13

krinsman:
This one contains the big migration from `tornado.options` to `traitlets`. I have spread it out over several commits to make the diffs slightly more readable, but I think the final commit "pulling the plug" and switching from one argument parsing system to the other will necessarily be somewhat long. I apologize in advance.


krinsman:
Didn't mean to close this, I deleted the branch and then created a new one with the same name, which had the effect of closing this one.

Anyway this PR is now out-of-date/redundant/etc. because of the following merged PR: https://github.com/jupyter/nbviewer/pull/913

PR #900, opened 2020-02-13
remove unused callback argument in fetch

minrk:
we have never used it and it is removed in tornado 6

since we are still passing callback=None, this is causing upstream errors not to be raised, since raise_errors is now the second positional argument and we are effectively passing `raise_errors=None`, which, being Falsy, suppresses upstream request errors.


minrk:
This is required for tornado 6 compatibility

PR #905, opened 2020-03-04
helm chart

minrk:
we need a helm chart to migrate to a new host, so this is a first go of it.


minrk:
This is now deployed on OVH and powering 50% of traffic to nbviewer.jupyter.org

PR #908, opened 2020-03-05
Merge pull request #908 from minrk/docker-lite

minrk:
main difference is to use wheels instead of conda

results in a much lighter image and much faster image builds

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,7 @@ build
 dist
 node_modules
 screenshots
+nbviewer/git_info.json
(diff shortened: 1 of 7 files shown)


minrk:
marked wip because the git info isn't working. We should stage this during build because right now, nbviewer *only* works when installed with `pip install -e`, not normal installs.

minrk:
Fixed the git_info issue by creating git_info.json, similar to what we do in IPython

PR #961, opened 2020-10-20
Remove Bokeh links

bryevdv:
cc @minrk 

Please let me know if there is anything else that needs to be done to remove Bokeh from the nbviewer.org site.


minrk:
That should do it, thanks!

minrk:
This is deployed. Will take a little while for caches to clear.

PR #968, opened 2020-11-05
switch master to default branch

shreddd:
This addresses the issue where the github repo/user URL in  nbviewer defaults to master. Instead we pull the value of default_branch from the GH API and and use that. 

Note that I didn't change the uri_rewrites because I don't think these are being used - but let me know if I missed something.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -3,7 +3,7 @@ repos:
   rev: v1.3.5
   hooks:
   - id: reorder-python-imports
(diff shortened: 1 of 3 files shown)


shreddd:
@krinsman  - this PR addresses the default_branch issue

minrk (approved):
Thanks! I'll look at deploying this in the morning

minrk:
This change has been deployed. Thanks!

PR #1002, opened 2022-03-15
Update URL, nbviewer is no longer a subdomain.

Carreau:
This was done for security reasons


Carreau:
Thanks, much appreciated.

Issue #630, opened 2016-09-30
Access S3 URLs directly


rgbkrk:
Similar to how the Github provider works, we should allow nbviewer to access S3 buckets and objects. If the runner of nbviewer has the appropriate access role, it could provide access to private notebooks as well.

adammclaurin:
This is a fantastic idea

jasonhao518:
Looking forward for this feature, potentially it can also support minio, since they are same protocol

PR #1009, opened 2022-05-19
Add s3 provider

dolfandringa:
This addresses #630 . It allows downloading notebooks from s3 buckets through the S3 uri (`s3://bucket/path/to/key`). If the file/bucket is public, this should work straight away. But if it is a private bucket, it can be configured by setting the [appropriate environmental variables](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/credentials.html#environment-variables) in the docker container or providing the [Shared credentials file](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/credentials.html#environment-variables) to the same user running NBViewer. 

I attempted to write a test too, but since the test server is started with `subprocess.POpen` we can't mock boto3. So that test has been commented for now. If there is a way to start the server without using subprocess, that would allow mocking, I could uncomment (and expand) that test.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,6 +46,11 @@ docker run -p 8080:8080 -e 'GITHUB_OAUTH_KEY=YOURKEY' \
 Or to use your GitHub personal access token, you can just set `GITHUB_API_TOKEN`.
 
 
+## S3 buckets
+Files in S3 buckets can be access by their s3 uri like `s3://bucket/path/to/key`. This works directly for public buckets. If you want to access private buckets, you need to provide the s3 authentication credentials to the docker container or in your environment. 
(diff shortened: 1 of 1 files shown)


nbviewer/providers/s3/handlers.py
```diff
@@ -0,0 +1,149 @@
...
+        """Download the notebook"""
+        headers = await self.get_notebook_headers(path)
+        filename = os.path.basename(path)
+        self.set_header("Content-Length", headers["ContentLength"])
+        # Escape commas to workaround Chrome issue with commas in download filenames
+        self.set_header(
+            "Content-Disposition",
+            "attachment; filename={};".format(filename.replace(",", "_")),
+        )
+        if self._downloaded_path == path and self._downloadable_data is not None:
```
dolfandringa, line 46: I wanted to prevent re-fetching the file from s3. Maybe there is a beter way to cache the boto3 s3 transfer using memcached somehow. Any suggestions are welcome.

PR #1012, opened 2022-06-30
fix #1011 -- Some content on the front page is VERY dated

### LINKED ISSUE #1011: Some content on the front page is VERY dated
I was just browsing the front page (https://nbviewer.org/), and the notebook link **right in the middle** (maybe the most central one that a person browsing might be drawn to) is very dated.  If you click and browse the notebooks, you just see they use Python 2.  If you actually look at the github repo itself -- https://github.com/ptwobrussell/Mining-the-Social-Web-2nd-Edition -- there is a huge warning at the top **HTTP 301: Don't Use This Repository - 17 Jan 2019**, and there is a recommendation to use https://github.com/mikhailklassen/Mining-the-Social-Web-3rd-Edition instead.  So the central main nbviewer link has been deprecated for 3.5 years.     I have no comments about other content, but when fixing this, it may be worth checking that the other content hasn't been deprecated for years...

image

williamstein:
fix #1011

This updates "Mining the Social Web" to 3rd edition, and also replaces the image with one from the third edition of the book (same dimensions). I also changed the title to be consistent with the other titles and more meaningful. I didn't run any tests.


krinsman:
This looks good in terms of substance -- there appears to be a style consistency error (that's what Black is checking for), but that might be a pre-existing issue with the head commit on the main branch (i.e. not introduced by this PR) -- I would have to check / look into it more closely.

Admittedly I have went somewhat AWOL on this project over the past two years while I was focused on my PhD research

If this goes unattended for say, two weeks, feel free to ping me again and I'll schedule/set aside time over a weekend for myself to review it more carefully -- I want to be much more confident that I correctly understand the current state of the project after being AWOL for so long

PR #959, opened 2020-10-04
Update code for nbconvert >= 6.0 (#958, #953) 

robindebois:
Hi,

Ran into the same problem as issue #958. I had to make several changes to get it to work; not all of them I'm fully confident with.

- had to replace `exporter_map` from nbconvert with `get_exporter` function..
- the `template_file` in HTMLExporter didn't roll...  so I updated it to `template_name`... 🤨.

Leaving out the second fix will result in the following error:

```
    Traceback (most recent call last):
      File "/usr/local/lib/python3.7/site-packages/nbviewer/providers/base.py", line 727, in finish_notebook
        self.config,
      File "/usr/local/lib/python3.7/concurrent/futures/thread.py", line 57, in run
        result = self.fn(*self.args, **self.kwargs)
      File "/usr/local/lib/python3.7/site-packages/nbviewer/render.py", line 56, in render_notebook
        html, resources = exporter.from_notebook_node(nb)
      File "/usr/local/lib/python3.7/site-packages/nbconvert/exporters/html.py", line 119, in from_notebook_node
        return super().from_notebook_node(nb, resources, **kw)
      File "/usr/local/lib/python3.7/site-packages/nbconvert/exporters/templateexporter.py", line 384, in from_notebook_node
        output = self.template.render(nb=nb_copy, resources=resources)
      File "/usr/local/lib/python3.7/site-packages/nbconvert/exporters/templateexporter.py", line 148, in template
        self._template_cached = self._load_template()
      File "/usr/local/lib/python3.7/site-packages/nbconvert/exporters/templateexporter.py", line 355, in _load_template
        return self.environment.get_template(template_file)
      File "/usr/local/lib/python3.7/site-packages/jinja2/environment.py", line 883, in get_template
        return self._load_template(name, self.make_globals(globals))
      File "/usr/local/lib/python3.7/site-packages/jinja2/environment.py", line 857, in _load_template
        template = self.loader.load(self, name, globals)
      File "/usr/local/lib/python3.7/site-packages/jinja2/loaders.py", line 429, in load
        raise TemplateNotFound(name)
    jinja2.exceptions.TemplateNotFound: basic
```

While I was waiting for the docker build I fixed 2 typos from #947 . 

I'm not sure if this is merge-worthy... feedback or a better solution would be appreciated.

[Edit]: also some help with solving all the weird 503 errors in the automated test is welcome. Not sure what I broke here..

PR #1015, opened 2022-08-11
Upgrade nbconvert

martinRenou:
Revive https://github.com/jupyter/nbviewer/pull/1003 and merge with https://github.com/jupyter/nbviewer/pull/1014

closes #1003 
closes #1014 
closes #959


minrk:
@martinRenou this mostly works now, but the lab template's font-awesome 5 is conflicting with the font-awesome 4 used on the nbviewer page, erasing the icons:

Screen Shot 2022-08-22 at 14 47 05
 
Not sure if there's a simple fix for that.

martinRenou:
Thanks a lot for pushing on this @minrk !

> but the lab template's font-awesome 5 is conflicting with the font-awesome 4 used on the nbviewer page, erasing the icons:

In Voila we do the following, maybe that will fix the nbviewer case:

```
<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/@fortawesome/fontawesome-free@^5/css/all.min.css" type="text/css" />
<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/@fortawesome/fontawesome-free@^5/css/v4-shims.min.css" type="text/css" />
```

I am not sure where to add this though. I wonder if this should go directly in nbconvert.

minrk:
We should update to fa5, but this backward-incompatibility is definitely an issue. The shims do work in layout.html (after our own css), but result in a different size than before:

before:
Screen Shot 2022-08-24 at 12 39 30

after:
Screen Shot 2022-08-24 at 12 39 27

Carreau:
The navbar seem to be missing visually as well.
Screen Shot 2022-10-31 at 09 31 59

nbviewer/templates/notebook.html
```diff
@@ -114,12 +114,9 @@
 
 
 {% block body %}
+    BREAKCRUMB
```
Carreau, line 117: DEBUG ?

Carreau:
A huge chink of the css issue seem to be with

```
  {% block ipywidgets %}
    <!-- block ipywidgets notebook.html -->
    <script>
      (function() {
        function addWidgetsRenderer() {
          var mimeElement = document.querySelector('script[type="application/vnd.jupyter.widget-view+json"]');
          var scriptElement = document.createElement('script');
          var widgetRendererSrc = '{{ ipywidgets_base_url }}@jupyter-widgets/html-manager@{{ jupyter_widgets_html_manager_version }}/dist/embed-amd.js';
          var widgetState;

          try {
            widgetState = mimeElement && JSON.parse(mimeElement.innerHTML);

            if (widgetState && (widgetState.version_major < 2 || !widgetState.version_major)) {
              widgetRendererSrc = '{{ ipywidgets_base_url }}jupyter-js-widgets@{{ jupyter_js_widgets_version }}/dist/embed.js';
            }
          } catch(e) {}

          scriptElement.src = widgetRendererSrc;
          document.body.appendChild(scriptElement);
        }

        document.addEventListener('DOMContentLoaded', addWidgetsRenderer);
      }());
    </script>
    <!-- end block ipywidgets notebook.html -->
  {% endblock ipywidgets %}
```

That dynamically inject css.

nbviewer/templates/notebook.html
minrk: yes, debugging

minrk:
@Carreau thanks for picking this up!

README.md
```diff
@@ -138,7 +138,7 @@ Pass `-d` or `--debug` to `invoke less` to create a CSS sourcemap, useful for de
 
 ```shell
 $ cd <path to repo>
-$ python -m nbviewer --debug --no-cache
+$ python -m nbviewer --debug --no-cache --host=127.0.0.1
```
Carreau, line 141: I think listening on 0.0.0.0 by default is a security problem. Well, here on macos, it present me from doing so without root access so It's ok-ish, but i'd like to fix it.

Carreau (changes requested):
Slides are also quite broken. 

For example you can compare with  https://nbviewer.org/format/slides/github/astrojuanlu/talk-dataframes/blob/main/slides.ipynb

Carreau:
Most of the reveal issue seem to be coming down to font-size being way to big. 

Something like the following in the right place make it bearable.
```
div.reveal {
    font-size: 10px;

}
```

Now we do seem to have 2 sets of reveal controls on the page if you look carefully: 

Screen Shot 2022-10-31 at 16 15 10

Carreau:
So there are indeed 2 reveal. One done by nbconvert, and one by nbviewer. 

Really this is crazy that we are embedding a full nbconvert with `body|safe`, that even include DOCTYPE ????. And of course the Reveal from nbconvert has no hooks, so we can't listen to the slide change event to hide the the nbviewer header.

The new nbconverts are really backward in term of reusing the templates...

Carreau (approved):

nbviewer/providers/github/handlers.py
```diff
@@ -155,6 +155,9 @@ async def get(self, user, repo):
         self.redirect(new_url)
 
 
+TEST_KEYS = {(1, 2, 3): ""}
+
+
```
Carreau, line 160:
```suggestion
```

nbviewer/templates/formats/slides.html
```diff
@@ -37,59 +37,4 @@
...
-          ]
-        });
-
-        var update = function(event){
-          window.scrollTo(0,0);
-          if(MathJax.Hub.getAllJax(Reveal.getCurrentSlide())){
-            MathJax.Hub.Rerender(Reveal.getCurrentSlide());
-          }
-          var idx = Reveal.getIndices();
-          $('#menubar').headroom((!idx.h && !idx.v) ? 'pin' : 'unpin');
```
Carreau, line 86: That is the part we can't do without changes to nbviewer.

Carreau (approved):
I think that's about as far as I we can reasonably take it in a single PR. It works, but there is still alot of work to do.

Carreau:
I think we should leapfrog nbconvert 6.x and go directly to 7.x, I'm going to create a 1.x branch that is the current main, and suggest main become 2.x.

minrk:
If that makes it easier, by all means!

Carreau:
> If that makes it easier, by all means!

It's more on the communication side, there will be likely so much change between 1.x and main, that it does make sens to bump major version number. This also let us re-do a 1.x at one point if ever necessary. 

I've push a 1.x branch, and I'm going to merge this and keep iterating, once this is merged, main should not be considered stable as a number of notebook likely won't render yet, and reveal is still partially broken .

martinRenou:
Thanks a lot for pushing on this!

mgeier:
Would it be possible to get another `nbconvert` upgrade on https://nbviewer.org/?

I think the latest release should fix an issue where SVG images are not displayed correctly: https://github.com/jupyter/nbconvert/issues/1863.
See also #1037.

PR #1022, opened 2022-10-23
Add shoutout to fastly for CDN

yuvipanda:


yuvipanda:
Me and @minrk moved us completely off cloudflare (except for some redirects, which shall be done soon)

minrk:
Noting that fastly has provided this for years. There's still a shoutout to Fastly in the footer on rendered pages.

Issue #664, opened 2017-01-20
got 400: bad request


tarokiritani:
My jupyter notebook renders ok on github, but not on nbviewer. I got error "400: Bad Request".

My notebook is here:

https://github.com/tarokiritani/in-vivo-wcr/blob/master/electrophysiology.ipynb

parente:
The notebook is stored in Git LFS. It's likely the current nbviewer implementation doesn't support fetching from that.

tarokiritani:
Yes, it is a very large (>100 MB) notebook. Thanks a lot for looking into the issue.

Honghe:
@parente Does nbviewer have plan to support Git LFS ipynb?

banesullivan:
+1 for LFS support.

This is needed with the recent support for rendering `panel` outputs

PR #1033, opened 2022-12-26
Add provider for huggingface.co: Very basic `uri_rewrites`

julien-c:
Hi!

I use https://nbviewer.org/ quite a bit and we are also considering implementing a notebook-rendering service on [huggingface.co](https://huggingface.co) that could potentially be built on top of nbconvert and/or nbviewer.

As a quick useful feature, I added URL rewrite from huggingface "blob" URLs, i.e. those you get when navigating to a ipynb file inside a repository, to the "resolve" URLs which we use to serve actual content.

This works for both:
- files stored in raw git: e.g. https://huggingface.co/pceiyos/fake_news_detection_nlp/blob/main/Fake_News_Classificaton.ipynb
- files stored in LFS, i.e. large notebooks over 10MB (also related to #664): https://huggingface.co/spaces/NimaBoscarino/climategan/blob/main/notebooks/plot_metrics.ipynb

Let me know if this looks ok!

### DIFF
--- a/README.md
+++ b/README.md
@@ -159,6 +159,7 @@ Providers are sources of notebooks and directories of notebooks and directories.
 - `url`
 - `gist`
 - `github`
+- `huggingface`
 - `local`
 
 #### Writing a new Provider
--- a/nbviewer/providers/__init__.py
+++ b/nbviewer/providers/__init__.py
@@ -11,7 +11,7 @@ default_providers = [
 
 default_rewrites = [
     "nbviewer.providers.{}".format(prov)
-    for prov in ["gist", "github", "dropbox", "url"]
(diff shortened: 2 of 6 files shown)


SylvainCorlay:
This looks good to me. I approved the CI to run and will merge if it turns out green!

julien-c:
Thanks for merging! will look for it when it's deployed.

SylvainCorlay:
> Thanks for merging! will look for it when it's deployed.

This is not done upon merging PRs (An explicit action must be taken).

SylvainCorlay:
@julien-c there you go!

Screenshot 2023-01-11 at 13 33 41

PR #1056, opened 2024-01-18
 Support  Google Colab and Mineo

eternalcrash:
This pull request aims to enhance the functionality of nbviewer by adding support for opening notebooks in Google Colab and MINEO. With this addition, users will have the flexibility to seamlessly view and interact with their notebooks in these environments directly from nbviewer.

Changes Made:
- Added integration for Google Colab (colab.research.google.com).
- Integrated support for Mineo (mineo.app) to open notebooks.
imagen


minrk:
Sorry for the delay, but I don't think we should link to these services from nbviewer.org. A customizeable "launch button(s)" config would be welcome, though.

PR #1067, opened 2024-11-11
Drop Python 3.8

trexfeathers:
Hi all, since [Python 3.8 is now end-of-life](https://devguide.python.org/versions/), I thought I would put up a PR that removes it from this repo. Thanks


minrk:
trying to bring this back to life a bit. Thank you!

PR #1069, opened 2024-12-12
Added note about how to use https feature to the FAQ

captainshar:
This wasn't obvious to me and I didn't see info about it on the site - others might find this detail useful too!

### DIFF
--- a/nbviewer/templates/faq.md
+++ b/nbviewer/templates/faq.md
@@ -166,7 +166,7 @@ nbviewer was generously hosted by Rackspace until March, 2020.
 
 ## Can I access nbviewer.org over https?
 
-Yes. Please do.
+Yes. Please do. Use the URL prefix `https://nbviewer.org/urls/` to access nbviewer.org over https. Note that the URL prefix uses `urls/` instead of `url/`.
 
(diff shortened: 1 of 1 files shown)


nbviewer/templates/faq.md
```diff
@@ -166,7 +166,7 @@ nbviewer was generously hosted by Rackspace until March, 2020.
 
 ## Can I access nbviewer.org over https?
 
-Yes. Please do.
+Yes. Please do. Use the URL prefix `https://nbviewer.org/urls/` to access nbviewer.org over https. Note that the URL prefix uses `urls/` instead of `url/`.
```
minrk, line 169: Thanks for the PR! This question is about accessing nbviewer.org itself, not nbviewer.org accessing _your notebooks_ over https, which is a slightly different question. Do you want to add a new question about that immediately below here? As it is, this addition gives the impression that `/urls/` is required to access nbviewer over https, when ~everything is over https already with the sole exception of accessing your own notebooks at `/url/`. I'd also suggest a recommendation pointing to typing URLs into the main page as the first thing folks should do, which already always produces the right URL for http and https links, as well as GitHub, etc.

PR #1084, opened 2025-09-20
Add jupytercon link to navbar

choldgraf:
This adds a link to jupytercon 2025 in the navbar. I couldn't figure out how to make a banner header, so I added it as an additional link. Hopefully it is minimal enough that it wouldn't be disruptive. Here's a screenshot of what it looks like:

CleanShot 2025-09-19 at 19 20 37@2x

### DIFF
--- a/nbviewer/templates/layout.html
+++ b/nbviewer/templates/layout.html
@@ -102,7 +102,7 @@
       <div class="collapse navbar-collapse">
         <ul class="nav navbar-nav navbar-right">
           <li>
-            <a class="active" href="http://jupytercon.org">🚀 JUPYTERCON 2025 · NOV 4-5</a>
+            <a class="active" href="http://jupytercon.com">🚀 JUPYTERCON 2025 · NOV 4-5</a>
           </li>
           <li>
             <a class="active" href="http://jupyter.org">JUPYTER</a>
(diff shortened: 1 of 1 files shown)


nbviewer/templates/layout.html
```diff
@@ -101,6 +101,9 @@
 
       <div class="collapse navbar-collapse">
         <ul class="nav navbar-nav navbar-right">
+          <li>
+            <a class="active" href="http://jupytercon.org">🚀 JUPYTERCON 2025 · NOV 4-5</a>
```
choldgraf, line 105:
```suggestion
            <a class="active" href="http://jupytercon.com">🚀 JUPYTERCON 2025 · NOV 4-5</a>
```

jasongrout:
@minrk - is there an easy way to customize just nbviewer.org?

minrk:
No worries! I don't think we need to roll it back, we can add the necessary hook to shift it to nbviewer.org-deploy and then move it at that point.

I'm trying to bring nbviewer back to life a bit this week, and getting the CD set up and updating readme, etc. so folks can help out and we can keep it up-to-date a bit more.

jasongrout:
> This adds the link to all private nbviewer deployments, not just nbviewer.org, is that what we want?

Ah, now it makes sense why you added plausible analytics in https://github.com/jupyter/nbviewer.org-deploy/pull/3, in the deploy repo, rather than here.

